PR #10, opened 2021-10-10
Created bigquery__test_unique to handle expressions passed to BQ uniq…

DigUpTheHatchet:


dbt/include/bigquery/macros/adapters.sql
```diff
@@ -194,3 +193,24 @@
   {%- endcall %}
 
 {% endmacro %}
+
+
+{% macro bigquery__test_unique(model, column_name) %}
```
DigUpTheHatchet, line 198: Is this the correct file for this macro to be to?

jtcohen6: Sure, this works!

It could also go into its own file, something like `dbt/include/bigquery/macros/generic_tests/unique.sql`. There's no functional difference, it's just a question of organization.

tests/integration/schema_tests_test/models-v2/bq-models/schema.yml
```diff
@@ -16,6 +16,9 @@ models:
   # this whole model should pass and run
   - name: table_summary
     description: "The summary table"
+    tests: 
+     - unique:
+          column_name: "concat(favorite_color_copy, count)"
```
jtcohen6, line 21: 👍

tests/integration/schema_tests_test/models-v2/bq-models/schema.yml
```diff
@@ -12,10 +12,18 @@ models:
       - name: favorite_color
         tests:
           - accepted_values: { values: ['blue', 'green'] }
+      - name: ephemeral_copy
+        description: "Testing a column with the same name as the model"
+        tests:
+          - not_null
+          - unique
```
DigUpTheHatchet, line 19: @jtcohen6 I added a test to cover ([#33](https://github.com/dbt-labs/dbt-bigquery/issues/33) as well

PR #48, opened 2021-10-29
Fix problem with row access policy doesnt return bytes proccessed.

### LINKED ISSUE #47: [BUG] Problem create table with row access policy
### Describe the bug
I create a row policy level in bigquery for the user who is used in dbt to access it, and when I try to run `dbt run -m + dim_table` the tables that is created using reference of the table with row access policy got a macro error. I got the run target for that table, and running directly on bigquery everything works fine.
I trying to use this process to add the dbt user in row policy access to spread more easily on others tables this row I don't want to show in Metabase.

### Steps To Reproduce
Create a row access policy using
```
CREATE OR REPLACE ROW ACCESS POLICY teste
ON `tableA`
GRANT TO ("serviceAccount:dbt_service_account@project_id.iam.gserviceaccount.com")
FILTER USING (id != "99");
```
Where service account is the same used previous in dbt to access.

then create a sql file with
```
select
    *
from
    {{ ref('tableA') }}
```
Without row access policy evething works fine, with row access policy i got and error

### Expected behavior
Expect the table be created and in table a when try to get information from id 99 will return an empty result
But instead i got and error

```
Completed with 1 error and 0 warnings:

Compilation Error in macro statement (macros/core.sql)
  bad operand type for abs(): 'NoneType'
  
  > in macro materialization_table_bigquery (macros/materializations/table.sql)
  > called by macro statement (macros/core.sql)

Done. PASS=8 WARN=0 ERROR=1 SKIP=1 TOTAL=10
```

### System information
I try in both
```
installed version: 0.19.1

Plugins:
  - redshift: 0.19.1
  - postgres: 0.19.1
  - bigquery: 0.19.1
  - snowflake: 0.19.1
```
and in 
```
installed version: 0.21.0

Plugins:
  - redshift: 0.21.0
  - postgres: 0.21.0
  - bigquery: 0.21.0
  - snowflake: 0.21.0
```

debian buster
python 3.8

imartynetz:
resolves #47 

### Description

Solution for the bug I found with row policy access

I found the problem and a solution for it.
Because of some security in row access policy, when the user that is include in row access policy run a query with the table, he receive None as how much bytes is proccessed
[https://cloud.google.com/bigquery/docs/best-practices-row-level-security](url)
So with that dbt the `bytes_processed`return None value and break the `abs(bytes_processed)`
### Checklist

- [ x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ x] I have run this code in development and it appears to resolve the stated issue
- [ x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-bigquery next" section.

### DIFF
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -19,7 +19,6 @@ from google.oauth2 import (
 )
 
 from dbt.adapters.bigquery import gcloud
-from dbt.utils import format_bytes, format_rows_number
 from dbt.clients import agate_helper
 from dbt.config.profile import INVALID_PROFILE_MESSAGE
 from dbt.tracking import active_user
@@ -215,6 +214,28 @@ class BigQueryConnectionManager(BaseConnectionManager):
     def commit(self):
         pass
 
+    def format_bytes(self, num_bytes):
+        if num_bytes:
+            for unit in ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB']:
+                if abs(num_bytes) < 1024.0:
+                    return f"{num_bytes:3.1f} {unit}"
+                num_bytes /= 1024.0
+
+            num_bytes *= 1024.0
+            return f"{num_bytes:3.1f} {unit}"
+
+        else:
+            return num_bytes
+
+    def format_rows_number(self, rows_number):
+        for unit in ['', 'k', 'm', 'b', 't']:
+            if abs(rows_number) < 1000.0:
+                return f"{rows_number:3.1f}{unit}".strip()
+            rows_number /= 1000.0
+
+        rows_number *= 1000.0
+        return f"{rows_number:3.1f}{unit}".strip()
+
     @classmethod
     def get_bigquery_credentials(cls, profile_credentials):
         method = profile_credentials.method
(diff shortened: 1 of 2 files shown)


dbt/adapters/bigquery/connections.py
```diff
@@ -380,7 +380,7 @@ def execute(
             query_table = client.get_table(query_job.destination)
             code = 'CREATE TABLE'
             num_rows = query_table.num_rows
-            bytes_processed = query_job.total_bytes_processed
+            bytes_processed = query_job.total_bytes_processed or 0
```
jtcohen6, line 383: I think it would make more sense for `total_bytes_processed` to be `None` here, rather than `0`. Per its type, it _can_ be `None`:
https://github.com/dbt-labs/dbt-bigquery/blob/d6369963a25fdfc188b205c20bc4191bbe49faa1/dbt/adapters/bigquery/connections.py#L83

A few points:
- `format_bytes` + `format_rows_number` are only used in `dbt-bigquery`, so it would make sense to move those methods out of `dbt-core` (`core/dbt/utils.py`) and into this package/repo
- We should update the logic here, or in those methods, to return `None` if `bytes_processed` is `None`, rather than trying to wrap it in `abs()` and raising an error

jtcohen6 (commented):
Thanks for the PR @imartynetz! Are you able to sign the CLA?

I left one comment on the implementation.

dbt/adapters/bigquery/connections.py
imartynetz: Hey @jtcohen6 I sign the CLA first with my personal account, than i figure out that my local github is linked with the company email, than I sign again with other email.

I make the change to be `0` to avoid change too much in the code and avoid to break in other part of the code, and it`s a safe approuch

imartynetz: Maybe change to this?
```
def format_bytes(num_bytes):
    if num_bytes:
        for unit in ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB']:
            if abs(num_bytes) < 1024.0:
                return f"{num_bytes:3.1f} {unit}"
            num_bytes /= 1024.0

        num_bytes *= 1024.0
        return f"{num_bytes:3.1f} {unit}"
    
    else:
        return None
```
What do you think?

jtcohen6: That makes sense to me!

imartynetz: I made a new change to add this methods to `bigquery/connections` so just need to be deleted from core.

jtcohen6: That sounds good to me. It's not a huge deal if that method remains unused in `core/utils.py` for the time being.

jtcohen6 (commented):
@imartynetz Do you think you'll be able to sign the CLA? Unfortunately, I won't be able to merge these changes otherwise

imartynetz:
@jtcohen6 CLA is that google forms right? I sign twice, with both email that my github is linked. Don't know whats happening.

jtcohen6:
@cla-bot check

jtcohen6:
Hooray! CLA check worked. I'm going to close and reopen just to trigger the integration tests, now that I've added the `ok to test` label

jtcohen6 (approved):
The only test failures are unrelated (schema version bumps). This change LGTM! Thanks so much @imartynetz!

imartynetz:
> The only test failures are unrelated (schema version bumps). This change LGTM! Thanks so much @imartynetz!

You're welcome, always nice to help.

imartynetz:
Hello @jtcohen6. Just to check, this hotfix will be commit to dbt-bigquery 0.21.0 to?

jtcohen6:
@imartynetz This fix will be included in v1.0.0 final. (It's already been included in v1.0.0rc1.) Unfortunately, we split out the adapter plugins into their own repos between v0.21 and v1.0, so it would require extra manual, error-prone work to backport fixes into the `dbt-core` repo for earlier versions.

Final release is planned for early December. That's not too long to wait, I hope!

imartynetz:
> @imartynetz This fix will be included in v1.0.0 final. (It's already been included in v1.0.0rc1.) Unfortunately, we split out the adapter plugins into their own repos between v0.21 and v1.0, so it would require extra manual, error-prone work to backport fixes into the `dbt-core` repo for earlier versions.
> 
> Final release is planned for early December. That's not too long to wait, I hope!

Nice to know, i will wait for this new version.

PR #56, opened 2021-11-05
Bump artifact versions in tests

jtcohen6:
Ditto https://github.com/dbt-labs/dbt-redshift/pull/35

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -157,6 +157,10 @@ jobs:
           pip install tox
           pip --version
           tox --version
+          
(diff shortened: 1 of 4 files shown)


jtcohen6:
CI check is failing because it's installing `dbt-bigquery` (latest) from source ahead of `dbt-core`. That has the effect of installing `dbt-core==1.0.0b2`, rather than `dbt-core` (latest) from source. This isn't an issue we've run into before, but it may come up again. I think the answer would be to install `dbt-core` as a separate prerequisite step.

PR #61, opened 2021-11-09
Add metrics to docs_generate test

jtcohen6:
I think I was too eager in merging #60


jtcohen6:
Jk, I was too eager in merging the py36 issue. I'm going to combine this commit with the one in #59

PR #63, opened 2021-11-10
Making scopes a configurable profile config option for BigQuery

### LINKED ISSUE #23: Tighten OAuth Scopes for BigQuery
### Describe the feature
Reduce the BigQuery OAuth scopes down to the minimal set needed, in both dbt and dbt Cloud.

### Additional context
Currently, the dbt BigQuery connector requests these three OAuth scopes:

*  'https://www.googleapis.com/auth/bigquery'
*  'https://www.googleapis.com/auth/cloud-platform'
*  'https://www.googleapis.com/auth/drive'

The BigQuery scope is needed to access the database, but the cloud-platform and drive scopes are probably too broad.  These scopes were originally added to address issue dbt-labs/dbt#502, primarily to allow for access for reading from Google Sheets.  However, I don't immediately see a need for the `cloud-platform` scope, which gives access to a wide range of GCP resources, such as the following:

*  View and manage your data across all Google Cloud Platform services, such as:
*  View and manage your tables, datasets, and jobs in Google BigQuery
*  View and manage your data in Google Cloud Storage
*  View and manage your instances in Google Cloud SQL

Similarly, the `drive` scope has this access:

```
This app wants permission to access everything in your Google Drive. It will be able to do the same things you can do, including:

See your files
Upload and download your files
Delete your files
See the names and emails of people you share files with
Share and stop sharing your files with others
Remove people from your files
Organize your Drive
There may be private information in your Google Drive, like financial records, medical reports, photos or tax info.
```

I would think that minimally, these scopes could be reduced to the 'read-only' variants, and could probably be reduced further depending on the access needed for external tables.  Maybe something like:

*  'https://www.googleapis.com/auth/cloud-platform.read-only'
*  'https://www.googleapis.com/auth/drive.readonly'

But I don't know yet whether these scopes are too restrictive.

Also note that dbt Cloud has the same list of OAuth scopes, so whatever is changed in dbt should also be changed in dbt Cloud.

bborysenko:
resolves #23 

### Description

Add possibility to configure custom scopes for Google Cloud OAuth down to the minimal set if needed. By default it too broad.

### Checklist

- [x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-bigquery next" section - https://github.com/dbt-labs/docs.getdbt.com/pull/914.

### DIFF
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -110,6 +110,12 @@ class BigQueryCredentials(Credentials):
     client_secret: Optional[str] = None
     token_uri: Optional[str] = None
 
+    scopes: Optional[Tuple[str, ...]] = (
+        'https://www.googleapis.com/auth/bigquery',
+        'https://www.googleapis.com/auth/cloud-platform',
+        'https://www.googleapis.com/auth/drive'
+    )
+
     _ALIASES = {
         'project': 'database',
         'dataset': 'schema',
@@ -147,10 +153,6 @@ class BigQueryCredentials(Credentials):
 
 class BigQueryConnectionManager(BaseConnectionManager):
     TYPE = 'bigquery'
-
-    SCOPE = ('https://www.googleapis.com/auth/bigquery',
-             'https://www.googleapis.com/auth/cloud-platform',
-             'https://www.googleapis.com/auth/drive')
 
     QUERY_TIMEOUT = 300
     RETRIES = 1
@@ -245,16 +247,16 @@ class BigQueryConnectionManager(BaseConnectionManager):
         creds = GoogleServiceAccountCredentials.Credentials
 
         if method == BigQueryConnectionMethod.OAUTH:
-            credentials, _ = get_bigquery_defaults(scopes=cls.SCOPE)
+            credentials, _ = get_bigquery_defaults(scopes=profile_credentials.scopes)
             return credentials
 
         elif method == BigQueryConnectionMethod.SERVICE_ACCOUNT:
             keyfile = profile_credentials.keyfile
-            return creds.from_service_account_file(keyfile, scopes=cls.SCOPE)
+            return creds.from_service_account_file(keyfile, scopes=profile_credentials.scopes)
 
         elif method == BigQueryConnectionMethod.SERVICE_ACCOUNT_JSON:
             details = profile_credentials.keyfile_json
-            return creds.from_service_account_info(details, scopes=cls.SCOPE)
+            return creds.from_service_account_info(details, scopes=profile_credentials.scopes)
 
         elif method == BigQueryConnectionMethod.OAUTH_SECRETS:
             return GoogleCredentials.Credentials(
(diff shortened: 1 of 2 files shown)


dbt/adapters/bigquery/connections.py
```diff
@@ -5,7 +5,7 @@
 from functools import lru_cache
 import agate
 from requests.exceptions import ConnectionError
-from typing import Optional, Any, Dict, Tuple
+from typing import Optional, Any, Dict, Tuple, List
```
apevzner-dp, line 8: Not necessary

dbt/adapters/bigquery/connections.py
```diff
@@ -110,6 +110,12 @@ class BigQueryCredentials(Credentials):
     client_secret: Optional[str] = None
     token_uri: Optional[str] = None
 
+    scopes: Optional[List[str]] = (
```
apevzner-dp, line 113: `Optional[List[str]]` → `Tuple[str, ...]`

See [doc](https://docs.python.org/3/library/typing.html#typing.Tuple).

dbt/adapters/bigquery/connections.py
```diff
@@ -240,21 +242,21 @@ def format_rows_number(self, rows_number):
         return f"{rows_number:3.1f}{unit}".strip()
 
     @classmethod
-    def get_bigquery_credentials(cls, profile_credentials):
+    def get_bigquery_credentials(self, profile_credentials):
```
apevzner-dp, line 245: Not necessary, you [have to pass](https://docs.python.org/3/library/functions.html#classmethod
) `cls` instead of `self` in a `@classmethod`.

apevzner-dp (changes requested):

jtcohen6 (commented):
@bborysenko This is looking good to me!

Have you been able to test it locally, and confirm that it's working? E.g. exclude `drive` from `scopes`, and check that a model selecting from a GSheet indeed raises an authentication error.

Could you please:
- Add an entry to the top of `CHANGELOG.md` in this repo
- Open a related issue in https://github.com/dbt-labs/docs.getdbt.com to add `scopes` to [`bigquery-profile.md`](https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/reference/warehouse-profiles/bigquery-profile.md) (and, if you're up to it, a PR to do it, targeting the `next` branch)

bborysenko:
@jtcohen6 Updated the changelog and created PR https://github.com/dbt-labs/docs.getdbt.com/pull/914 for docs. 

I'm not able to test it locally for the next few days. @apevzner-dp Can you please take care of it and confirm that it's working as expected? Thanks!

apevzner-dp:
@bborysenko @jtcohen6 I have tested the changes locally and **I confirm that it's working**.

I've created a BigQuery table with a GSheets doc as an external source and after I've created a dbt model which makes a query to this table. Then, I played with the new `scopes` configuration in the profile. When the `https://www.googleapis.com/auth/drive` scope was specified in the profile, the model worked well, while when I removed this scope from the profile, the `Permissions denied` error occurred.

jtcohen6 (approved):
Thanks for the contribution @bborysenko!

+ @apevzner-dp for the assist!

PR #109, opened 2022-01-26
Make sure partition_by is in config if setting require_partition_filter

### LINKED ISSUE #107: Make require_partition_filter=true in dbt_project.yml affect only partitioned tables
### Describe the feature

Currently if I add `require_partition_filter=true` to my dbt_project.yml it will apply this to all models in my repo, including the models that materialize as views or non-partitioned tables when materialized. This will throw the following error in BigQuery:

```
require_partition_filter value in table OPTIONS is only supported for a CREATE TABLE statement with a PARTITION BY clause
```

I would like a feature that lets me in my dbt_project.yml to be able to only apply the `require_partition_filter` to models that will materialize as partitioned tables. 

One alternative is to go through every single model `config` in my repo and only add `require_partition_filter` to the model configs that materialize as partitioned tables. However we have hundreds of models so this would be a very manual and cumbersome process.

### Additional context
N/A

### Who will this benefit?
This will benefit users that want to add `require_partition_filter` universally and directly in their dbt_project.yml and have it affect only the right models instead of manually going through all their configs. 

Based on difficulty / timeline for making this feature available, I could be willing to help

oliverrmaa:
resolves #107

### Description

This PR adds an additional condition in `get_table_options` for `require_partition_filter`. Essentially, both the `require_partition_filter` and `partition_by` options must both not be none for `require_partition_filter` to work. This way, non-partitioned tables cannot have `require_partition_filter` applied to them.

This should allow users to just put `require_partition_filter` in the dbt_project.yml file instead of needing to make sure they only add it to configs for models that has `partition_by` in it. 

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/impl.py
+++ b/dbt/adapters/bigquery/impl.py
@@ -767,7 +767,7 @@ class BigQueryAdapter(BaseAdapter):
             # It doesn't apply the `require_partition_filter` option for a temporary table
             # so that we avoid the error by not specifying a partition with a temporary table
             # in the incremental model.
-            if config.get('require_partition_filter') is not None:
+            if config.get('require_partition_filter') and config.get('partition_by') is not None:
                 opts['require_partition_filter'] = config.get(
                     'require_partition_filter')
             if config.get('partition_expiration_days') is not None:
--- a/dbt/adapters/bigquery/impl.py
+++ b/dbt/adapters/bigquery/impl.py
@@ -767,7 +767,8 @@ class BigQueryAdapter(BaseAdapter):
             # It doesn't apply the `require_partition_filter` option for a temporary table
             # so that we avoid the error by not specifying a partition with a temporary table
             # in the incremental model.
-            if config.get('require_partition_filter') and config.get('partition_by') is not None:
+            if config.get('require_partition_filter') is not None and \
+                    config.get('partition_by') is not None:
                 opts['require_partition_filter'] = config.get(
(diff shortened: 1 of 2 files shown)


dbt/adapters/bigquery/impl.py
```diff
@@ -767,7 +767,7 @@ def get_table_options(
             # It doesn't apply the `require_partition_filter` option for a temporary table
             # so that we avoid the error by not specifying a partition with a temporary table
             # in the incremental model.
-            if config.get('require_partition_filter') is not None:
+            if config.get('require_partition_filter') and config.get('partition_by') is not None:
```
hui-zheng, line 770: I  think maybe we shall state out two notNone checks? 
```suggestion
            if config.get('require_partition_filter') is not None and config.get('partition_by') is not None:
```

oliverrmaa: I think both should do the same, but I can definitely be more explicit if we need! Will wait for other opinions on this as well

McKnight-42:
@oliverrmaa seems you may be behind `dbt-bigquery` if you could update your local branch by pulling down the latest changes and re-push should get the other tests to trigger and keeps your local upto date incase you think of something you'd like to add to this later.

oliverrmaa:
@McKnight-42 , okay no problem, I have updated the branch to be up to date with `dbt-bigquery`

McKnight-42 (approved):

Issue #75, opened 2021-11-28
Support for ingestion time partition table on BigQuery as incremental materialization


github-christophe-oudar:
### Describe the feature

This is a follow-up from a discussion over [the topic](https://getdbt.slack.com/archives/C50NEBJGG/p1636997196076000) with @jtcohen6.

#### The issues with current incremental materialization implementation

Ingestion time partition tables are supported as `table` materialization right now but the support is meant to be deprecated.
The specificity for those tables is that partitioning field is `_PARTITIONTIME` OR `_PARTITIONDATE` as a "pseudo column".
Since that column doesn't really exist as a column within the table, Google doesn't allow the same kind of operations on the table.
Indeed if you would like to use

```
partition_by={
 "field": "_PARTITIONTIME",
 "data_type": "timestamp"
}
```

BigQuery doesn't let you create a ingestion time partitioned table using usual dbt approach (https://cloud.google.com/bigquery/docs/creating-partitioned-tables#create_an_ingestion-time_partitioned_table):
```
CREATE TABLE
  mydataset.newtable (transaction_id INT64)
PARTITION BY
  DATE_TRUNC(_PARTITIONTIME, DAY)
AS (
SELECT TIMESTAMP("2021-11-01") as _PARTITIONTIME, 1 as transaction_id
)
```
will fail as _PARTITIONTIME is not directly in the defined columns. 

---
#### Working approach

So the required approach is to have first:
```
CREATE TABLE
  mydataset.newtable (transaction_id INT64)
PARTITION BY
  DATE_TRUNC(_PARTITIONTIME, DAY)
```  

And then:

```
INSERT INTO mydataset.newtable (_PARTITIONTIME, transaction_id)
SELECT TIMESTAMP("2021-11-01"), 1 
```

Once we move to the merge part, it's indeed possible to insert data with

```
merge into mydataset.newtable as DBT_INTERNAL_DEST
    using (
    SELECT TIMESTAMP("2021-11-03") as _PARTITIONTIME, 3 as transaction_id
    ) as DBT_INTERNAL_SOURCE
    on FALSE

when not matched by source
        and DBT_INTERNAL_DEST._PARTITIONTIME in unnest([TIMESTAMP("2021-11-03")])
    then delete

when not matched then insert
    (`_PARTITIONTIME`, `transaction_id`)
values
    (`_PARTITIONTIME`, `transaction_id`)
```

Practically, it won't be 100% straightforward as the it requires to use
[...]

PR #136, opened 2022-03-13
Support for ingestion time partition table on BigQuery as incremental materialization

Kayrnt:
resolves https://github.com/dbt-labs/dbt-bigquery/issues/75

### Description

Support for ingestion time partition table on BigQuery as incremental materialization

### Checklist


dbt/include/bigquery/macros/materializations/incremental.sql
```diff
@@ -27,6 +27,84 @@
...
+        where false
+        limit 0
+    {% endcall %}
+    {%- set result = load_result('get_columns_with_types_in_query') -%}
+    {{ log('result ' ~ result) }}
+    {%- set table = result.table -%}
+    {{ log('table ' ~ table) }}
+    {%- set columns = result.columns -%}
+    {{ log('columns ' ~ columns) }}
+    {{ return(load_result('get_columns_with_types_in_query').table.columns | list) }}
```
github-christophe-oudar, line 44: Hello @jtcohen6 @McKnight-42, I could use some help over here:
To create a table such as 
```
create or replace table `project`.`cou`.`test_ingestion_dbt` (`x` INT64)
```

Yet, as it goes through agate lib apparently, I'm getting
```
create or replace table `project`.`cou`.`test_ingestion_dbt` (`x` <dbt.clients.agate_helper.Number object at 0x104688ee0>)
```

I guess I need to use `get_columns_in_relation` from the adapter so that I get the real BQ type and not a `Number` which can't be used for creating the table.
My intent is to use the `data_type` here https://github.com/dbt-labs/dbt-bigquery/pull/136/files#diff-167e3557df7f18f1520c5db0045dfac9923e38a617c909d703be844192b28ebeR76

Except for that, I'm making progress but we're not there yet (once I have everything working, I'll still need to write tests).
I'm wasting a lot of time working logging fields for dev as I don't think I can work with a step by step debugger (in Python code and in Jinja), isn't it?

Thanks!

dataders:
@Kayrnt you're very close here. correct me if I'm wrong, but my understanding is that you'd like to:
> for a given relation (table/view), give me the column type for each column.

The way to do this is with column_type column of the table returned by the `get_catalog()` ([`bigquery__get_catalog()`](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/include/bigquery/macros/catalog.sql#L2-L209)) macro.

Not only do I totally understand the approach you're taken, but I've also made the same mistake a few times. Your `call statement('get_columns_with_types_in_query'....` construction was sound, and it will return an empty table with column names.

The problem is with `load_result` which takes the json returned by the `call` statement and converts it into an Agate table. When Agate is given data to turn into an Agate table, there's a specific way it guesses the type of each columnn provided by looking at the values in each column. Because there are no example values for each column, Agate defaults to assuming they are `Number` type. If you're interested, you can read more on [Agate Data Types](https://agate.readthedocs.io/en/latest/api/data_types.html),  [Agate's type inference](https://agate.readthedocs.io/en/latest/api/type_tester.html), and how dbt-core modifies the Agate's type inference and even adds a new type in [`agate_helper.py`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/clients/agate_helper.py).


However, you are correct that directly querying BQ via an API will return the column types, however, AFAICT, there's nothing in dbt-core or dbt-biqquery that uses the metadata in every query response and passes the needed types to Agate before loading. Perhaps it could be though starting with [`BigQueryConnectionManager.get_table_from_response()`](https://github.com/dbt-labs/dbt-bigquery/blob/edebc4728fe6fc6cac79d853387060142e7b9e06/dbt/adapters/bigquery/connections.py#L358-L361). Agree or disagree: @VersusFacit @McKnight-42 @jtcohen6?

dbt/adapters/bigquery/impl.py
```diff
@@ -209,14 +223,22 @@ def check_schema_exists(self, database: str, schema: str) -> bool:
...
             return self._get_dbt_columns_from_bq_table(table)
 
         except (ValueError, google.cloud.exceptions.NotFound) as e:
             logger.debug("get_columns_in_relation error: {}".format(e))
             return []
 
+    @available.parse(lambda *a, **k: [])
+    def add_time_ingestion_partition_column(self, columns) -> List[BigQueryColumn]:
+        "Add time ingestion partition column to columns list"
+        columns.append(self.Column("_PARTITIONTIME", "TIMESTAMP", None, None))
```
dataders, line 239:
```suggestion
        columns.append(self.Column("_PARTITIONTIME", "TIMESTAMP", None, "NULLABLE"))
```

There are four arguments to `Column.__init__` besides `self`: `column`, `dtype`, `fields`, and `mode`. I don't think that `mode` can be `None`, so either pass nothing or pass `"nullable"` I think.

https://github.com/dbt-labs/dbt-bigquery/blob/3d6986984fffb21d82d3365e7cc5d72545f602e6/dbt/adapters/bigquery/column.py#L23-L29

dbt/adapters/bigquery/connections.py
```diff
@@ -473,9 +476,14 @@ def execute(
...
             message = f"{code} ({num_rows_formated} rows, {processed_bytes} processed)"
+            fields = query_table.schema
 
         response = BigQueryAdapterResponse(  # type: ignore[call-arg]
-            _message=message, rows_affected=num_rows, code=code, bytes_processed=bytes_processed
+            _message=message,
+            rows_affected=num_rows,
+            code=code,
+            bytes_processed=bytes_processed,
+            fields=fields,
```
dataders, line 486: looks like you're introducing non-black formatted code. any chance you want to take formatting `imply.py` with `black`, to make your actual changes easier to diff here?

dataders (changes requested):
I think you're almost there!

dbt/adapters/bigquery/impl.py
github-christophe-oudar: Ok good point, I'll try tomorrow, my issue was on an uncommitted change but it's very similar 👍

dbt/adapters/bigquery/connections.py
github-christophe-oudar: Indeed I ran `black` after my changes but I definitely run it in a dedicated commit to push before mine.

jtcohen6:
Finally got a chance to look in slightly more detail. Very cool work!

I wonder if, rather than futzing with the addition of `fields()` to _all_ AdapterResponse objects, what we really want is a BigQuery-specific implementation of `get_columns_in_query`. This would be comparable to [the default implementation](https://github.com/dbt-labs/dbt-core/blob/8e9702cec50bfd41e8ce26f1f624be81e9cefba9/core/dbt/include/global_project/macros/adapters/columns.sql#L24), but using a similar underlying approach to `get_columns_in_relation` (i.e. leveraging the Python client).

Something like:
```sql
{% macro bigquery__get_columns_in_query(select_sql) %}
  {% set sql %}
    select * from (
      {{ select_sql }}
    ) as __dbt_sbq
    where false
    limit 0
  {% endset %}
  {{ return(adapter.get_columns_in_select_sql(sql)) }}
{% endmacro %}
```
Leveraging a new adapter method like:
```python
    @available.parse(lambda *a, **k: False)
    def get_columns_in_select_sql(self, select_sql: str) -> List[BigQueryColumn]:
        try:
            conn = self.connections.get_thread_connection()
            client = conn.handle
            query_job, iterator = self.connections.raw_execute(select_sql)
            query_table = client.get_table(query_job.destination)
            return self._get_dbt_columns_from_bq_table(query_table)

        except (ValueError, google.cloud.exceptions.NotFound) as e:
            logger.debug("get_columns_in_query error: {}".format(e))
            return []
```

Following the integration test example, that returns for me:
```sql
  {%- set columns = get_columns_in_query(sql) -%}
  {%- do log(columns, info = true) -%}
```
```
[<BigQueryColumn id (INT64, NULLABLE)>, <BigQueryColumn field1 (STRING, NULLABLE)>, <BigQueryColumn field2 (STRING, NULLABLE)>]
```

github-christophe-oudar:
Thanks a lot for the feedback & the details @jtcohen6, I'll try that!
Putting it in the `AdapterResponse` felt like it was a good way to provide the information for other usages but since it brings serialization issues you're probably right about isolating that usage.

McKnight-42:
@Kayrnt fantastic work on this! looks like we've developed a conflict in some files if you can update to get tests rerunning that would be fantastic. if you can't please let me know.

github-christophe-oudar:
Hello @McKnight-42, thank you for having a look. I can't commit on a timeline till the very end of May as explained in my latest Slack message:
https://getdbt.slack.com/archives/C030A0UF5LM/p1651225949330499?thread_ts=1648824631.462679&cid=C030A0UF5LM

Feel free to take over if you want 👍
I would love to see the features shipped in the next release!

McKnight-42:
@Kayrnt alright I will look into clearing up the conflicts and see where we can get it as time allows. feel free to hop back onto when you can and stay involved!!!!

github-christophe-oudar:
I can comment and review a bit but next days will be busy on my end so expect some delay. Good luck on conflicts, it's a good opportunity to deeply review the code 😉

github-christophe-oudar:
Hello, how is it going on your end? I'm still traveling till the end of May but I'm looking forward to making progress on that PR to discuss #167 :+1:

jtcohen6:
Hey @Kayrnt! J'espère que tu profites bien des vacances. Sorry for the lack of response on our end these past few weeks.

As part of our work for v1.2, we're planning to meaningfully refactor the incremental materialization, starting with incremental strategies (https://github.com/dbt-labs/dbt-core/issues/5245). The code here has gotten a bit out of hand, and it's become quite tricky for us to maintain, especially as we onboard new folks to help out.

We also rolled out a new more intuitive testing framework in v1.1, which uses more standard `pytest` builtins. We're converting all our old integration tests to use the new framework. In any place where we're adding new functionality, and extending old-school test cases to test it, we're taking it as an opportunity to first convert those tests.

The capabilities added here and in #167 are ones I'd _very_ much like to support, ideally in `dbt-bigquery==1.2.0`. I'm immensely grateful for all your work in these issues and PRs, all the way to working reference implementations! After discussing with the Core engineering team, we think the right move is to "freeze" new additions to the incremental materializations, in `dbt-core` and in our adapter plugins, while we pursue that refactor. After the code is cleaner and the interfaces are clearer, I have a lot of appetite to help get this functionality merged in. That may look like helping you rebase these changes, if you have the availability/stamina next month. If you don't (which is 100% reasonable), one of us can adapt these PRs, and take them over the finish line.

github-christophe-oudar:
Hello @jtcohen6, tout se passe bien de mon côté mis à part pour tous les coups de soleil que j'ai reçu !

Ok no wonder, you're busy I had a quick look at your core roadmap and there is definitely a lot of work to ship everything!
Thank you for all the details you're sharing with the community!
I'll have a proper read once I get back as I'm curious about some plans but it's another topic.
Anyway I understand there is a lot of refactoring to be done to be able to move forward confidently and scale your team.
Hopefully it's also an opportunity for the dbt labs team to write down a more comprehensive documentation for contributors to understand the high level structure, the development environment and CI process if you find time to. It is already nice but I think I could have avoided you some questions and save a lot of time with some more details.

I'll likely be able to support the PRs over the next months (hopefully it won't be as much work as the original implementation).
In the meantime, I might use my fork in my organization as we're maintaining 2 systems because of those missing features. It might be an opportunity to find out some flaws till it's available to all users.

I'm following #dbt-core-development on Slack, let me know if there is another more appropriate channel to keep up with the changes. Let me know if I can help a bit (or provide feedback if you need some).

github-christophe-oudar:
I've rebased the change to work with the python.
The code should be properly tested and tests appears to pass.
I'll have few tests locally to confirm it works as intended but you can rereview the code.

Kayrnt:
What's the status on your end @jtcohen6 @McKnight-42? It would be sad to miss the 1.3 release 😉
@dataders I think you'll have to dismiss your review because it looks like I can't (and I think I made the change since then, but feel free to review 🙌 ). Thanks!

dbt/adapters/bigquery/impl.py
```diff
@@ -69,6 +73,13 @@ def render(self, alias: Optional[str] = None):
         else:
             return f"{self.data_type}_trunc({column}, {self.granularity})"
 
+    def render_wrapped(self, alias: Optional[str] = None):
+        """Wrap the partitioning column when time involved to ensure it is properly casted to matching time."""
+        if self.data_type in ("date", "timestamp", "datetime"):
```
colin-rogers-dbt, line 78: do we need to include "time" type here as well?

colin-rogers-dbt: ahh makes sense, thanks for the clarification!

McKnight-42:
@github-christophe-oudar shined a light on this one while testing locally as part of review looking good, hoping to get it merged soon!

github-christophe-oudar:
It looks like the stars are about to align then! ⭐ ⭐ ⭐ 
After that I'll rebase https://github.com/dbt-labs/dbt-bigquery/pull/167
Let's be ambitious and try to merge both in September 🙌

github-christophe-oudar:
👋 
Any chance we merge over the next weeks?
It will become a blocker on my end, so I'm wondering if I need to fork for my organisation.
Thanks 🙇

colin-rogers-dbt (approved):
LGTM

colin-rogers-dbt:
@cla-bot check

PR #142, opened 2022-03-16
Implement initial adapter tests on bigquery

### LINKED ISSUE #135: [CT-362] Initial use of new adapter testing framework
We need to pull in the new adapter test framework, and get the basic tests working.

gshank:
resolves #135

### Description

Initial implementation of tests ported from dbt-adapter-tests

### Checklist

### DIFF
--- a/tests/functional/adapter/test_basic.py
+++ b/tests/functional/adapter/test_basic.py
@@ -1,0 +1,52 @@
+import pytest
+
+from dbt.tests.adapter.basic.test_base import BaseSimpleMaterializations
+from dbt.tests.adapter.basic.test_singular_tests import BaseSingularTests
+from dbt.tests.adapter.basic.test_singular_tests_ephemeral import (
+    BaseSingularTestsEphemeral,
+)
+from dbt.tests.adapter.basic.test_empty import BaseEmpty
+from dbt.tests.adapter.basic.test_ephemeral import BaseEphemeral
+from dbt.tests.adapter.basic.test_incremental import BaseIncremental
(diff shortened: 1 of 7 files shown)


tests/functional/adapter/test_basic.py
```diff
@@ -0,0 +1,54 @@
...
+)
+from dbt.tests.adapter.basic.test_empty import BaseEmpty
+from dbt.tests.adapter.basic.test_ephemeral import BaseEphemeral
+from dbt.tests.adapter.basic.test_incremental import BaseIncremental
+from dbt.tests.adapter.basic.test_generic_tests import BaseGenericTests
+from dbt.tests.adapter.basic.test_snapshot_check_cols import BaseSnapshotCheckCols
+from dbt.tests.adapter.basic.test_snapshot_timestamp import BaseSnapshotTimestamp
+
+"""
+Trying to create view `dbt-dev-168022`.`test16473926894805161392_test_basic`.`swappable`,
```
gshank, line 16: @jtcohen6 What is your take on how to avoid this error?

jtcohen6: @gshank This is a "good" error! And it doesn't look like we have any existing integration tests to confirm that this "good" error is raised in this case, so I'd also call this a "good" test case :)

I think we could take either/both of these approaches:
- xfail the standard test with this error message — it looks 
- inherit + reimplement the standard test, adding in `--full-refresh`, so that it can succeed

Maybe both, if that's straightforward?

ChenyuLInx (approved):
LGTM

PR #121, opened 2022-02-13
Adding upload_file adapter

pgoslatara:
resolves #102

### Description

Adding an `upload_file` adapter that uses the [load_table_from_file](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.load_table_from_file) method. This adapter takes multiple arguments allowing maximum customisation of the [LoadJobConfig](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.LoadJobConfig.html#google.cloud.bigquery.job.LoadJobConfig) class.

Example files:
- ./samples/three_columns.csv:
```
id,name,created_at
1,john,2022-02-06 19:24:23.614 UTC
```
- ./samples/id_only.csv:
```
id
1
```
- ./samples/users.ndjson:
```
{"id":1,"name":"Alice"}
{"id":2,"name":"Bob"}
{"id":3,"name":"Carol"}
```

How to use:
- Via command line:
```
dbt -d run-operation bigquery__upload_file --args "{local_file_path: './samples/three_columns.csv', database: '<GCP_PROJECT>', table_schema: 'dbt', table_name: 'three_columns', skip_leading_rows: 1, autodetect: True, write_disposition: 'WRITE_TRUNCATE'}"

dbt -d run-operation bigquery__upload_file --args "{local_file_path: './samples/id_only.csv', database: '<GCP_PROJECT>', table_schema: 'dbt', table_name: 'upload_id', skip_leading_rows: 1, write_disposition: 'WRITE_TRUNCATE', schema: '[{\"name\": \"employee_id\", \"type\": \"INTEGER\"}]'}"

dbt -d run-operation bigquery__upload_file --args "{local_file_path: './samples/users.ndjson', database: '<GCP_PROJECT>', table_schema: 'dbt', table_name: 'users', autodetect: True, write_disposition: 'WRITE_TRUNCATE', source_format: 'NEWLINE_DELIMITED_JSON'}"
```
- Via a pre_hook:
```
{{
    config(
        pre_hook=[
            "{{ bigquery__upload_file(
                './samples/three_columns.csv',
                model['database'],
                model['schema'],
                'three_columns',
                autodetect=True,
                skip_leading_rows=1,
                write_disposition='WRITE_TRUNCATE',
                source_format='CSV'
            ) }}",
            "{{ bigquery__upload_file(
                './samples/id_only.csv',
                model['database'],
                model['schema'],
                'id_only',
[...]

PR #153, opened 2022-04-03
Adding upload_run_results macro

### LINKED ISSUE #152: [CT-451] upload_file macro does not work for run_results.json
### Describe the bug
The `upload_file` macro does not work with `run_results.json` as the file does not conform to BigQuery's accepted file formats as it can contain empty dict's as values, for example "... "env": {} ...".

### Steps To Reproduce
`dbt -d run-operation upload_file --args "{local_file_path: './target/run_results.json', database: '<GCP_PROJECT>', table_schema: '<DATASET>', table_name: '<TABLE_NAME>', autodetect: True, write_disposition: 'WRITE_TRUNCATE', source_format: 'NEWLINE_DELIMITED_JSON'}"`

### Expected behavior
A single row is added to the specified table.

If applicable, add screenshots or log output to help explain your problem.

### System information
```
installed version: 1.1.0-b1
   latest version: 1.0.4

Your version of dbt is ahead of the latest release!

Plugins:
```

Ubuntu 20.04

Python 3.9.10

pgoslatara:
resolves #152

### Description

Adding a new macro, `upload_run_results`, that piggybacks the `upload_file` macro to upload the `run_results.json` file produced by dbt runs. As previously mentioned in #121 the file does not conform to BigQuery's accepted file formats as it can contain empty dict's as values, for example "... "env": {} ...". The new macro automatically infers the schema of the `run_results.json` file and adds an "\_\_unknown\_\_" column to allow uploading.

Notes:
* `run_results.json` contains environment variables that start with "DBT_ENV_CUSTOM_ENV_", [source](https://docs.getdbt.com/reference/dbt-jinja-functions/env_var#custom-metadata), however any sensitive variables are recommended to start with "DBT_ENV_SECRET_", [source](https://docs.getdbt.com/reference/dbt-jinja-functions/env_var#secrets). As such it does not make sense for this macro to filter sensitive entries prior to uploading.
* The default write_disposition of the load_table_from_file method is "WRITE_APPEND", [source](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.WriteDisposition.html#google.cloud.bigquery.job.WriteDisposition), hence using the `upload_run_results` macro repeatedly with default arguments will not result in data being overwritten.

### Checklist


McKnight-42:
@pgoslatara if you could please add a changelog entry and make sure your up to date with main i'm curious as to why the code check test didn't trigger quite yet, will test this locally and review once we have all tests passing. but its looking very good so far.

McKnight-42:
@pgoslatara if you could please add a changelog entry and make sure your up to date with main i'm curious as to why the code check test didn't trigger quite yet, will test this locally and review once we have all tests passing. but its looking very good so far.

dbt/adapters/bigquery/impl.py
```diff
@@ -709,6 +710,35 @@ def upload_file(self, local_file_path: str, database: str, table_schema: str,
         with self.connections.exception_handler("LOAD TABLE"):
             self.poll_until_job_completes(job, timeout)
 
+    @available.parse_none
+    def upload_run_results(self, local_file_path: str, database: str, table_schema: str,
```
jtcohen6, line 714: I'd like to see this method be more generic in one way, _and_ more narrowly targeted in another.
- More generic: Could we rename to just `upload_json_artifact`? Is there a reason why this only works for `run_results.json`, and wouldn't work for `manifest.json`, `catalog.json`, or `sources.json`?
- More narrowly targeted: The intent with this method is _not_ to support arbitrary file &rarr; table creation. It's very specifically to hack support for uploading metadata artifacts, and I'd like to keep it as narrowly targeted as possible. I'm not sure of the best way to do that—perhaps we could validate that the local file is named `manifest.json`, `catalog.json`, `run_results.json`, or `sources.json`?

To that end, I'd also propose we also make the even-more-generic `upload_file` _unavailable_ to the Jinja context — i.e. remove the `@available.parse_none`, and remove the `bigquery__upload_file` macro. (It's still good to have the separate method, as a way of factoring the code here.)

setup.py
```diff
@@ -74,6 +74,7 @@ def _get_dbt_core_version():
         'google-cloud-bigquery>=1.25.0,<3',
         'google-api-core>=1.16.0,<3',
         'googleapis-common-protos>=1.6.0,<2',
+        'bigquery_schema_generator==1.5'
```
jtcohen6, line 77: Do we need to pin this to a specific version?

dbt/include/bigquery/macros/adapters.sql
```diff
@@ -222,3 +222,11 @@ having count(*) > 1
   {% do adapter.upload_file(local_file_path, database, table_schema, table_name, kwargs=kwargs) %}
 
 {% endmacro %}
+
+{% macro bigquery__upload_run_results(local_file_path, database, table_schema, table_name) %}
+
+  {{ log("kwargs: " ~ kwargs) }}
```
jtcohen6, line 228: Let's move the log message up into the `upload_run_results` (Python method), where it's slightly easier to construct a more descriptive message, and pass into the `logger` (= BigQuery `AdapterLogger`)

jtcohen6 (changes requested):
@pgoslatara Thanks for the contribution! Both of the notes you've included (secrets + write_append) make sense to me.

I know I missed the original PR (#121), so apologies for taking this opportunity to revisit some of the logic added there as well. From my point of view, I very much want to support the intended use case (https://github.com/brooklyn-data/dbt_artifacts/issues/6), while also keeping guardrails in place to prevent anti-pattern behavior. Happy to hear your thoughts as well

McKnight-42:
@pgoslatara if you could update your local to be in parity with main as it is, seems we need to add the pre-commit functionality to get code passing the local and  the code-quality tests. we are adding to all the repos. sorry for the inconvenience this may cause some changes to formatting of local files.

setup.py
pgoslatara: Done! 1df7b45e23337c2ab12c474f5b43f78003a9db32

dbt/include/bigquery/macros/adapters.sql
pgoslatara: Done! 32afa0eb243cd3f529a8c73f1fee066e152c0215

I've chosen "debug" rather than "info" however the print out still contains a large JSON string for the schema which can be offputting. Would be nice if there was an elegant way to only print this when the "-d" flag is passed.

dbt/adapters/bigquery/impl.py
pgoslatara: Thanks for the input @jeremy. Some thoughts on the latest commits I've pushed:
* Furthering your thoughts, if the desire is for a macro that uploads json artifacts and provides no other functionality then the file name selection is redundant. Instead a directory can be specified with the artifact file names hard-coded on the `dbt-bigquery` side. Does this make sense on your side, is there any way to create artifacts with custom names that I'm not aware of?
* BigQuery does not allow "." or "-" characters in column names. By default these characters will be replaced with double underscores ("__") although the option to select your own replacement character(s) is provided.
* BigQuery does not like lists inside lists as present with the "ref" key in `manifest.json`. For example, this file cannot be uploaded to Bigquery:
    ```
    {"col1": 1, "col2": [[1, 2]]}
    ```
    As it returns the error message "Nested arrays not allowed". In contrast this file can be uploaded:
    ```
    {"col1": 1, "col2": [1, 2]}
    ```
    The `upload_json_artifacts` method handles this for `manifest.json` by replacing the outer list with the value of the inner list. I'm not familiar with the values that can be contained in the "ref" key, do you see any issue with this approach?

pgoslatara:
@jtcohen6 @McKnight-42 Gentle reminder on this. Would love to get your thoughts on the latest commits and get this macro added to the package.

pgoslatara:
@jtcohen6 @McKnight-42 Another reminder on this, would love to get your input on the updated approach! Happy to bring this branch up to date with main once we're happy with the approach.

NiallRees:
Excited to see this added!

azhard:
Love this as well! Coming from someone who would like to use this functionality, it'd be great if you could specify what artifacts you want to upload vs. uploading all of the ones found in the directory.

pgoslatara:
@jtcohen6 @McKnight-42 Third time's a charm! Any chance I could get some input on these changes and whether I am on the right track or not?

jtcohen6:
@yu-iskw Agreed! I think that approach has the potential to be less wonky and error-prone, versus trying to teach dbt adapter plugins how to upload JSON files from the local file system — even for BigQuery, which has relatively solid mechanisms in place to do it

Fleid:
Hi @pgoslatara!

Would you say that [dbt_artifacts](https://github.com/brooklyn-data/dbt_artifacts) now supporting BigQuery solves your initial need to upload `run_results.json`?
If there's a gap, do you think that dbt_artifacts would be a better place to make that change?

I know I'm very very late to that discussion, sorry about that, I just want to make sure we're not missing something by letting the topic go stale.

Fleid:
Having slept on it, I'm going to close this PR.

The issue we are trying to solve is to upload dbt run metadata to the warehouse. This is outside of the scope of a specific adapter.
To solve that now, either use a package ([dbt-artifacts](https://github.com/brooklyn-data/dbt_artifacts), script it by using the warehouse SDK/CLI.
Longer term, I have a feeling that could be achieved from core, via the work being done on API-Ification and structured logging. I'll keep an eye on that.

PR #159, opened 2022-04-14
Set job_execution_timeout_seconds default to None

### LINKED ISSUE #158: [v1.1 regression] Remove default value for `job_execution_timeout_seconds`
### A brief history of `timeout_seconds`

Back in https://github.com/dbt-labs/dbt-core/pull/1963, which originally added timeout and query retry logic to `dbt-bigquery` (back in Nov 2019), it was proposed to add a universal `timeout_seconds` of 300s for all queries. At that point, Drew said:

> I don't think we actually want to enforce the query_timeout here -- the default is 300s which will cause a lot of BigQuery projects to start failing for no reason. I don't really buy that per-model timeouts are a good idea, and I don't think I'd be in favor of implementing them for other databases that dbt supports. Instead, I think timeouts like these are better handled by orchestration tools at the level of a dbt invocation. If there is sufficient interest in supporting per-model timeouts, I'd rather support it via a model config and not a profile config. The timeout_seconds config as it exists today is pretty heavy-handed! So, I'd be in favor of retaining the previous behavior, inconsistent as it may be.

Then in https://github.com/dbt-labs/dbt-bigquery/issues/20:

> We should consider supporting a dbt-wide timeout_seconds config in the future (this would apply to any query invoked by dbt), so let's continue to support the config in profiles.yml, but render a warning at run-start indicating that the config is not used.

> I can also imagine node-level timeouts which are better suited for model-level configs, eg:
> ```
> models:
>  timeout_seconds: 3600
> ```
>
> Here, if any model takes more than 3600 seconds to complete, it will be cancelled by dbt. You could equivalently do something > like:
> ```
> snapshots:
>  timeout_seconds: 3600
>```

so you could have different node-level semantics for different node types.

After we removed ingestion-time partitioned tables for v1.0 (#42), `timeout_seconds` had a default of 300 seconds, and zero effect whatsoever (since it only applied to those queries).

In https://github.com/dbt-labs/dbt-bigquery/pull/50, this config was revived, renamed `job_execution_timeout_seconds`, and it's now applied universally to _all queries_, including the execution of every dbt model's transformation SQL.

After putting all the pieces together, I now realize that will cause a **lot** of BigQuery models to start failing as soon as they upgrade to v1.1 — by my latest estimate, some 3k distinct models across 450 distinct projects.

### Next steps

Let's do two things:
- Right now: disable `job_execution_timeout_seconds` by default, rather than setting the default to 300 seconds. This should have no actual effect relative to v1.0 behavior, since this config was truly unused.
- Later: open an issue for the ability to override timeout/retry configs on a per-model level

jtcohen6:
Meta notes:
- resolves #158, which is a likely regression in v1.1.0rc1
- will require backporting to `1.1.latest`, and probably a `1.1.0rc2` release
- will require an update to docs: https://docs.getdbt.com/reference/warehouse-profiles/bigquery-profile#job_execution_timeout_seconds

### Description

This has the effect of passing in `timeout=None` to `query_job.result` by default (rather than 300 seconds). Per [BigQuery's docs](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJob.html#google.cloud.bigquery.job.QueryJob.result), this is an optional value, with default `None`:

> The number of seconds to wait for the underlying HTTP transport before using retry. If multiple requests are made under the hood, timeout applies to each individual request.

This restores parity with v1.0 behavior, where we didn't pass a `timeout` into `_query_and_results` for _anything_ except (undocumented) copy jobs. Obviously, if the user-provided value _is_ configured, we'll use it everywhere.

I saw two other places where we were previously using `timeout_seconds` to plumb our own logic:
- the expiration time for service account impersonation
- the polling time for uploading table contents (seeds)

In both cases, I just added back 300 seconds as the fallback value if `job_execution_timeout_seconds` isn't set by the end user. This should also restore parity with default behavior in v1.0, while still allowing end users to configure their own custom timeout.

### Checklist

- ~This PR includes tests, or tests are not required/relevant for this PR~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-## dbt-bigquery 1.1.0 (TBD)
+## dbt-bigquery 1.1.0 (Release TBD)
+
+### Fixes
+- Restore default behavior for query timeout. Set default `job_execution_timeout` to `None` by default. Keep 300 seconds as query timeout where previously used.
+
+## dbt-bigquery 1.1.0rc1 (April 13, 2022)
 
 ### Under the hood
 - Use dbt.tests.adapter.basic in tests (new test framework) ([#135](https://github.com/dbt-labs/dbt-bigquery/issues/135), [#142](https://github.com/dbt-labs/dbt-bigquery/pull/142))
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -103,7 +103,7 @@ class BigQueryCredentials(Credentials):
     job_retry_deadline_seconds: Optional[int] = None
     job_retries: Optional[int] = 1
     job_creation_timeout_seconds: Optional[int] = None
-    job_execution_timeout_seconds: Optional[int] = 300
+    job_execution_timeout_seconds: Optional[int] = None
 
     # Keyfile json creds
     keyfile: Optional[str] = None
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -301,7 +301,7 @@ class BigQueryConnectionManager(BaseConnectionManager):
             source_credentials=source_credentials,
             target_principal=profile_credentials.impersonate_service_account,
             target_scopes=list(profile_credentials.scopes),
-            lifetime=profile_credentials.job_execution_timeout_seconds,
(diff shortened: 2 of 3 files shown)


leahwicz (approved):
Good catch Jeremy!

PR #161, opened 2022-04-15
fix in dev_requirements reference in contributing.md

McKnight-42:
### Description

changes dev-requirements.txt -> dev_requirements.txt

### Checklist


kwigley:
do you think it makes sense to align with dbt-core's naming and change the file name and all references to it? It is named dev-requirements.txt over there https://github.com/dbt-labs/dbt-core/blob/main/dev-requirements.txt

McKnight-42:
I think the all the adapters if i'm not mistaken are on dev_requirements so either way swapping ref in contributing.md or changing the file name seems about same amount of pr's technically but if going to be fully aligned namespace wise is a gain yeah that might be better choice.

kwigley (changes requested):
request to rename file instead of updating docs here

leahwicz:
@McKnight-42 can we close this out then since we went with renaming?

PR #167, opened 2022-04-24
Optimize insert_overwrite incremental strategy with WRITE_TRUNCATE / Partition copy

### LINKED ISSUE #77: Optimize `insert_overwrite` incremental strategy with WRITE_TRUNCATE / Partition copy
### Describe the feature

This is a follow-up from a discussion over the topic with @jtcohen6.

Let's assume we have a model with incremental materialization and we're going for an incremental run.

```
merge into mydataset.newtable as DBT_INTERNAL_DEST
    using (
    SELECT * from mydataset.newtable__dbt_tmp
    ) as DBT_INTERNAL_SOURCE
    on FALSE

when not matched by source
        and DBT_INTERNAL_DEST._PARTITIONTIME in unnest([TIMESTAMP("2021-11-30")])
    then delete

when not matched then insert
    (`_PARTITIONTIME`, `transaction_id`)
values
    (`_PARTITIONTIME`, `transaction_id`)
```

However the merge will first delete the data to insert the data.
On big partitions, it can take a while resulting in a long operation compared to other approaches such as:
- select whole partition and insert in the destination table with a partition decorator (ie `mydataset.newtable$20211130`) while using `WRITE_TRUNCATE` setting.
- copying the whole partition ie: `bq cp mydataset.newtable__dbt_tmp$20211130 mydataset.newtable$20211130`.

As far as I know the fastest approach is to use `bq cp` (it might require to benchmark it though) unless you don't need a temporary table (ie no column change & single partition) in which case, you would use the input query and write to the destination table with the partition decorator directly.

The main inconvenient for that approach is that even though the `insert_overwrite` strategy operation isn't fully atomic (the temporary table is created and stays even if the `MERGE` fails), the `MERGE` query is atomic on all partitions.
So to keep the same behavior, it requires to have a single partition or accept the tradeoff of breaking the atomicity of the partitions replacement step.

Therefore it could be relevant to have a `copy_partitions` config to activate that approach.

It could also be a whole new incremental strategy if relevant.

### Additional context
I had a production case that I described in the thread where, using a single partition, `MERGE` version (insert_overwrite) was taking 43 minutes while the query with WRITE_TRUNCATE took 26 minutes for the same result.
At Teads, our internal BigQuery query wrapper tool is using a select & write_truncate since we're not using a temporary table as an intermediate step because we only process a single partition per query (which is a specific case compared to the dbt approach). It's quite a deal breaker to use dbt as is for those queries because of that performance overhead.
Of course, it would be much better if Google could make a server side optimisation on `MERGE` queries when it detects that pattern.

### Who will this benefit?
It will benefit anyone using `insert_overwrite`  incremental strategy with large partitions where delete/insert are long to process.

Yes

github-christophe-oudar:
resolves #77

### Description

- Add a `copy_partitions` to `partition_by` object (default to false), throw an error if `incremental_strategy != 'insert_overwrite'`.
- If `insert_overwrite` & `copy_partitions`, we replace the `MERGE` part by:
  - In static mode (partitions defined), we copy explicit partitions from the output in temp table
  - In dynamic mode, we retrieve the partitions from the output in temp table and copy all of them

### Checklist


dbt/adapters/bigquery/impl.py
```diff
@@ -47,11 +47,16 @@ class PartitionConfig(dbtClassMixin):
     data_type: str = "date"
     granularity: str = "day"
     range: Optional[Dict[str, Any]] = None
+    time_ingestion_partitioning: bool = False
+    copy_partitions: bool = False
```
colin-rogers-dbt, line 51: Is there a use case for copy_partitions outside of time_ingestion_partitioning or for not using copy_partitions with time_ingestion_partitioning? Just wondering if we need another config parameter

colin-rogers-dbt (commented):
Doing some local testing, one question so far but overall it seems to function well. Hope to be able to move this and #136 forward this week

Also: apologies for the extreme delay on this, still getting ramped up but we do want to prioritize getting this merged post-Coalesce

dbt/adapters/bigquery/impl.py
github-christophe-oudar: Yes, absolutely, it's also much faster than a merge so I'm going to use it on all my models whether it's time ingestion partitioning or column type partitioning! It really fits my use case because I'm usually processing models partition per partition but it might not fit everyone (though honestly, most people in my opinion that can afford a temporary loss of consistency for a speed gain 👍 )

colin-rogers-dbt: Is there a use case for `time_ingestion_partitioning = True` and `copy_partitions = False `? Should we allow users to do this?

@Fleid - I know you had concerns with adding more variables for users to set but I think we need to keep these separate

github-christophe-oudar: Yes, here some use cases:
`time_ingestion_partitioning = True` and `copy_partitions = False` => for atomic merge for time ingestion  partitioned table (useful for people that can't afford potential inconsistency)
`time_ingestion_partitioning = False` and `copy_partitions = True` => for fast atomic insert overwrite on a single partition or fast non atomic on multiple partitions (quite common at my company)
`time_ingestion_partitioning = False` and `copy_partitions = False` => before the 2 PRs
`time_ingestion_partitioning = True` and `copy_partitions = True` => for fast atomic insert overwrite on a single partition or fast non atomic on multiple partitions (a common use case at my company)

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
 - Use dbt.tests.adapter.basic in tests (new test framework) ([#135](https://github.com/dbt-labs/dbt-bigquery/issues/135), [#142](https://github.com/dbt-labs/dbt-bigquery/pull/142))
 - Adding pre-commit and black formatter hooks ([#147](https://github.com/dbt-labs/dbt-bigquery/pull/147))
 - Adding pre-commit code changes ([#148](https://github.com/dbt-labs/dbt-bigquery/pull/148))
+- Add support for ingestion time partitioned table using incremental materialization ([#136](https://github.com/dbt-labs/dbt-bigquery/pull/136))
```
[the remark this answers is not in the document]
github-christophe-oudar: Yes I'll update, it was done before changie was introduced

dbt/adapters/bigquery/impl.py
colin-rogers-dbt: Great context! I'm going to do a follow up PR after we merge to update our documentation with this context.

github-christophe-oudar:
@colin-rogers-dbt I rebased it, I think you have put the label to trigger the integration tests so that we can check there's no regression there too.

dbt/include/bigquery/macros/materializations/incremental_strategy/insert_overwrite.sql
```diff
@@ -9,18 +9,44 @@
...
-    tmp_relation, target_relation, sql, unique_key, partition_by, partitions, dest_columns, tmp_relation_exists
+    tmp_relation, target_relation, sql, unique_key, partition_by, partitions, dest_columns, tmp_relation_exists, copy_partitions
 ) %}
-
   {% if partitions is not none and partitions != [] %} {# static #}
+      {{ bq_static_insert_overwrite(tmp_relation, target_relation, sql, partition_by, partitions, dest_columns, copy_partitions) }}
+  {% else %} {# dynamic #}
+      {{ bq_dynamic_insert_overwrite(tmp_relation, target_relation, sql, unique_key, partition_by, dest_columns, tmp_relation_exists, copy_partitions) }}
+  {% endif %}
+
```
colin-rogers-dbt, line 46: Missing an an endmacro: 
```suggestion
{% endmacro %}
```

github-christophe-oudar: Well spotted, I struggled a bit with the rebase so I'm not surprised I missed something 🙈 
Thanks!

colin-rogers-dbt:
This integ test is failing due to a known issue with our testing. No action required from @Kayrnt

colin-rogers-dbt (approved):
LGTM

PR #177, opened 2022-05-01
Adding Python 3.10 to test matrix 

leahwicz:
### Description
This change started out as Python 3.10 support but then grew some. With this change, we now will have:
- Python 3.10 tests running for CI
- mypy upgraded to match Core
- mypy enabled for the repo
- third party stubs (needed for mypy)

### Checklist


gshank:
You can't add an __init__.py file underneath the 'dbt' directory. It breaks the plugin architecture.

dbt/__init__.py
```diff
@@ -0,0 +1,3 @@
+from pkgutil import extend_path
+
+__path__ = extend_path(__path__, __name__)
```
VersusFacit, line 3: I've read changing __path__ this way is uncommon. What's the reason here? Super curious about the use case.

VersusFacit (commented):
Question: In a previous effort, I played around with notions of mypy's added stub feature. I opted away from this after having several questions. Questions that come to mind:
* where all the agate stubs you've included came from!
* How do we purpose to keep this up to date with the changing library specification?
* What's the update procedure?

dbt/__init__.py
leahwicz: We had a convo [in Slack](https://dbt-labs.slack.com/archives/G01GRE7RQSF/p1651772259078419) here about it but the TLDR is that we reuse `dbt` namespace twice which confuses mypy. This helps just combine them for mypy to reference. In the long run, we should refactor the codebase to not have this issue

leahwicz:
Good questions Mila! 

> where all the agate stubs you've included came from!
I copied them from `dbt-core` where they already exist. 

> How do we purpose to keep this up to date with the changing library specification? / What's the update procedure?
In the majority of the cases, we should use the associated types package for each of these dependencies. I removed the `response` stubs I originally created myself and switched the `types-requests` package instead. This is the way we can keep up to date with dependencies over time. For dependencies that don't provide the associated types package, we don't have a good plan in place at the moment which is a gap. Agate is a dependency in general we want to remove in general. I'll take a look at the stubs we have in all the repos and start thinking if there is a better approach there. We can start a thread with the team too since I'm curious if others have seen this done in a certain way in the past

McKnight-42 (approved):
Looks good to me, not seeing any red-flags sticking out in the types

PR #183, opened 2022-05-05
Try using SQL for `create_schema`

### LINKED ISSUE #182: Prefer SQL for `create_schema` + `drop_schema`
reopening https://github.com/dbt-labs/dbt-bigquery/issues/30

Today, `dbt-bigquery` operates as a "base" (Python) adapter for most low-level operations. This means that standard stuff like creating schemas, dropping schemas, getting columns in relations, etc, are all actually API calls wired through the Python client.

As BigQuery adds more support for SQL, it's preferable to use it where possible, and where there are few trade-offs. Two reasons:
- It's clearer for end users to reason about, and read in the logs
- It moves `create_schema` + `drop_schema` into "user-space" code, allowing for custom reimplementations if desired. This is a capability users have on other databases. It's especially important for `create_schema` + `drop_schema`, since these don't pass through SQL/Jinja land _at all_ when called as `adapter` methods from dbt's Python tasks, namely [here](https://github.com/dbt-labs/dbt-core/blob/788694ec5b9796d45592a8e704a3618881775e2a/core/dbt/task/runnable.py#L540)

The trade-offs I can imagine:
- Performance. Generally, BigQuery's SQL "API" is always slower than its Python client API for the equivalent operation (and occasionally more expensive)
- Reliability. `get_columns_in_relation` can return a contracted `SchemaField`, which we can turn into dbt `BigQueryColumn` objects, much more easily than passing query results through `agate`

jtcohen6:
resolves #182

I'd like to avoid duplicating code from `SQLAdapter` if we can help it, but I'm not sure of the right way to plumb through its methods without turning `BigQueryAdapter` into an unholy hybrid of `BaseAdapter` + `SQLAdapter`.

### Checklist


jtcohen6:
- Unit tests for `create_schema` + `drop_schema` fail, which isn't too surprising
- All our legacy integration tests (`tests/integration`) are failing as well. Just out of curiosity, I'm going to reorder to run the new functional tests (`tests/functional`) first, since that was working for me locally

jtcohen6:
Cool! Per [latest GHA run](https://github.com/dbt-labs/dbt-bigquery/runs/6302960363?check_suite_focus=true), this change seems to be working with our new testing framework, just not the legacy one. I'm going to close this PR for now, and consider it blocked by our ongoing test conversion work.

jtcohen6:
The main value here is around converting `create_schema`, since that's what dbt runs at start-up. The only cases where dbt actually uses its own `drop_schema` method/macro are:
- clean up after integration/functional test
- clean up PR schemas in dbt Cloud

I'm running into a very confusing permissions issue at the intersection of the old testing framework and SQL `drop schema`. So, we could unblock this work for the meantime by moving ahead with _just_ `create schema`, and opening a new tech debt ticket to track the unfinished work around `drop schema`. I believe that work will be unblocked once we've totally put to rest the old testing framework.

McKnight-42:
This is very interesting, do you think we will eventually move completely away from using BaseAdapter for BigQuery?

github-christophe-oudar:
I would add that it's indeed pretty frustrating that we can't rely only on SQL on BigQuery. https://github.com/dbt-labs/dbt-bigquery/pull/167 is yet another example on how BigQuery lacks consistency on that topic and makes it harder for its users to rely only on SQL (and that we need some smart custom dbt logic to make it easier to work with for dbt users until Google figures it out).

tests/unit/test_bigquery_adapter.py
```diff
@@ -349,10 +349,10 @@ def test_get_relation(self):
...
         self.mock_connection_manager.get_bq_table.assert_called_once_with('db', 'schema', 'my_model')
 
-    def test_create_schema(self):
-        relation = BigQueryRelation.create(database='db', schema='schema')
-        self.adapter.create_schema(relation)
-        self.mock_connection_manager.create_dataset.assert_called_once_with('db', 'schema')
+#    def test_create_schema(self):
+#        relation = BigQueryRelation.create(database='db', schema='schema')
+#        self.adapter.create_schema(relation)
+#        self.mock_connection_manager.create_dataset.assert_called_once_with('db', 'schema')
```
McKnight-42, line 355: @jtcohen6 do we plan on leaving the test commented out or is it okay to remove?

jtcohen6: Good catch - probably okay to remove. I don't think we can have a dedicated unit test like this (mocking a Python method) if dbt is running real SQL instead. I also don't feel like we need a dedicated test for `create_schema`, since it's a required "setup" step in every single functional test.

Issue #180, opened 2022-05-04
[CT-604] `get_columns_in_relation` when database is missing


jtcohen6:
[Slack thread](https://getdbt.slack.com/archives/C37J8BQEL/p1651673844092349?thread_ts=1651242161.526509&cid=C37J8BQEL)

A highly specific edge case for code that worked in v1.0, but not in v1.1.

### Reproduction case

```sql
-- macros/test_get_cols.sql

{% test get_cols_in(model) %}

  {# The step which causes the issue #}
  {%- set relation = api.Relation.create(schema=model.schema, identifier=model.table) if execute -%}

  {% set columns = adapter.get_columns_in_relation(relation) %}
  
  select
    {% for col in columns %}
      {{ col.name }} {{ "," if not loop.last }}
    {% endfor %}
  
  from {{ model }}
  limit 0

{% endtest %}
```
```sql
-- models/my_model.sql
select 1 as id, 'text' as another_col
```
```yml
# models/whatever.yml
version: 2
models:
  - name: my_model
    tests:
      - get_cols_in
```

When I run this code locally, I see this error in the debug logs (`logs/dbt.log`):
```
BigQuery adapter: get_columns_in_relation error: Pass a string for project
```

Indeed, it looks like this error is because the `project`/`database` is empty when `relation` is created. When I update that line of code to be:
```
{%- set relation = api.Relation.create(database=model.database, schema=model.schema, identifier=model.table) if execute -%}
```
It works just fine.

So the question is, how was this working previously?

### Root cause

The origin of this change: https://github.com/dbt-labs/dbt-bigquery/pull/98

[`get_columns_in_relation`](https://github.com/dbt-labs/dbt-bigquery/blob/3bbd76b328f25ba919a99d40c08eed37fc8c1b28/dbt/adapters/bigquery/impl.py#L209) calls [`get_bq_table`](https://github.com/dbt-labs/dbt-bigquery/blob/3bbd76b328f25ba919a99d40c08eed37fc8c1b28/dbt/adapters/bigquery/connections.py#L555):

https://github.com/dbt-labs/dbt-bigquery/blob/8679a8bd7613b82da2a3032b15cc02973eb5a3b1/dbt/adapters/bigquery/connections.py#L550-L559

Starting in v1.1, these use a new mechanism (`DatasetReference` + `TableReference`), which expect all of `database`/`project`, `schema`/`dataset`, and `table_name` to be present.

[...]

PR #188, opened 2022-05-18
Add regression test case

McKnight-42:
#180 

### Description
checks for backwards compatibility in an edge case that database/schema information can be grabbed from default

### Checklist


tests/functional/test_get_columns_incomplete_database.py
```diff
@@ -0,0 +1,50 @@
...
+      - get_cols_in
+"""
+
+class TestIncompleteRelation:
+    @pytest.fixture(scope="class")
+    def properties(self):
+        return {"properties_model_yml.yml": properties__model_yml}
+
+    @pytest.fixture(scope="class")
+    def tests(self):
```
jtcohen6, line 42: Generic test definitions can go into either the `macros/` directory, or the `tests/generic` subdirectory. It doesn't look like the framework lets us specify the file path as `"generic/get_col_in.sql"`, so macros will do.
```suggestion
    def macros(self):
```
https://docs.getdbt.com/docs/guides/writing-custom-generic-tests

tests/functional/test_get_columns_incomplete_database.py
```diff
@@ -0,0 +1,50 @@
...
+        return {"tests__get_col_in.sql": tests__get_cols_in_sql}
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return { "models__my_model.sql": models__my_model }
```
jtcohen6, line 47: This model file needs to be named `"my_model.sql"`, so that it matches up with the tests defined on `my_model` in `properties__model_yml`:
```suggestion
        return { "my_model.sql": models__my_model }
```

tests/functional/test_get_columns_incomplete_database.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def test_incomplete_relation(self, project):
+        run_dbt(["run"])
```
jtcohen6, line 50:
```suggestion
        run_dbt(["run"])
        run_dbt(["test"])
```
Or just:
```suggestion
        run_dbt(["build"])
```

jtcohen6 (commented):
Looks good, nearly there! Remember you'll want to include the fix as well when backporting to `1.1.latest`.
https://github.com/dbt-labs/dbt-bigquery/blob/8aaa6ab6da182aa8e831f631feb12af5ec76d9ce/dbt/adapters/bigquery/connections.py#L558-L560

tests/functional/test_get_columns_incomplete_database.py
McKnight-42: do we need to augment the test name as well in that case? or do we have detection to remove the prefix for tests?

tests/functional/test_get_columns_incomplete_database.py
McKnight-42: made suggested changes @jtcohen6

McKnight-42:
the two docs_generate tests that are failing pass locally looking into issue in CI

PR #207, opened 2022-06-28
[bug] Fix error when attempting to log errors with --log-format json

### LINKED ISSUE #206: [CT-784] When using --log-format json, exceptions during query execution result in JSON Serialization errors
### Describe the bug
When an Exception is raised during query execution, and the log format is set to json, a JSON serialization exception will be raised instead of the original Exception.

### Steps To Reproduce

Create a dbt project with the following model:

```
select * from foo.bar
```

and run `dbt --log-format json run`

### Expected behavior

The error message should be something along the lines of:
`404 Not found: Table my-dataset:foo.bar was not found in location US"`

The actual error message (once parsed out of the "msg" field of the json logs: `dbt --log-format-json run | jq .msg`):

```
"Completed with 1 error and 0 warnings:"
"Compilation Error in macro statement (macros/etc/statement.sql)"
"  Object of type NotFound is not JSON serializable"
"  "
"  > in macro materialization_table_bigquery (macros/materializations/table.sql)"
"  > called by macro statement (macros/etc/statement.sql)"
```

### System information
```
Core:
  - installed: 1.1.1
  - latest:    1.1.1 - Up to date!

Plugins:
  - bigquery: 1.1.0 - Up to date!
```

macOS Catalina

`Python 3.8.7`

### Additional context

This is happening because of this line: https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/connections.py#L217

With the default log formatter, this Exception will be automatically stringified. The json log formatter instead just takes the raw exception and attempts to convert it to JSON.

OwenKephart:
resolves #206

### Description

Instead of attempting to log the bare Exception object, log the stringified version of it. This will have no impact on the default logging behavior, while allowing this message to be json serializable.

The contributing guide mentions using `changie` for changelog entries, but it doesn't seem like that's set up for this subrepo, let me know if I'm missing something!

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -215,8 +215,8 @@ class BigQueryConnectionManager(BaseConnectionManager):
         except Exception as e:
             exc_message = str(e)
             logger.debug("Unhandled error while running:\n{}".format(sql))
-            logger.debug(exc_message)
             if isinstance(e, RuntimeException):
+                logger.debug(exc_message)
                 # during a sql query, an internal to dbt exception was raised.
                 # this sounds a lot like a signal handler and probably has
                 # useful information, so raise it without modification.
@@ -225,6 +225,7 @@ class BigQueryConnectionManager(BaseConnectionManager):
             # don't want to log. Hopefully they never change this!
             if BQ_QUERY_JOB_SPLIT in exc_message:
                 exc_message = exc_message.split(BQ_QUERY_JOB_SPLIT)[0].strip()
+            logger.debug(exc_message)
             raise RuntimeException(exc_message)
 
     def cancel_open(self) -> None:
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -213,14 +213,14 @@ class BigQueryConnectionManager(BaseConnectionManager):
             raise RuntimeException(message)
 
         except Exception as e:
+            exc_message = str(e)
             logger.debug("Unhandled error while running:\n{}".format(sql))
-            logger.debug(e)
+            logger.debug(exc_message)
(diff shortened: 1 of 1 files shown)


dbt/adapters/bigquery/connections.py
```diff
@@ -213,18 +213,19 @@ def exception_handler(self, sql):
...
                 # during a sql query, an internal to dbt exception was raised.
                 # this sounds a lot like a signal handler and probably has
                 # useful information, so raise it without modification.
                 raise
-            exc_message = str(e)
             # the google bigquery library likes to add the query log, which we
             # don't want to log. Hopefully they never change this!
             if BQ_QUERY_JOB_SPLIT in exc_message:
                 exc_message = exc_message.split(BQ_QUERY_JOB_SPLIT)[0].strip()
+            logger.debug(exc_message)
```
dataders, line 228: does this need to happen within the `if BQ_QUERY_JOB_SPLIT in exc_message:` block?
```suggestion
                logger.debug(exc_message)
```

OwenKephart: @dataders I don't have strong feelings here, but in the original implementation, the error message would be logged in all scenarios (as `logger.debug(e)` was outside of all conditionals). In the proposed version, the error message would not be logged in the scenario that the Exception was not a RuntimeException, and BQ_QUERY_JOB_SPLIT was not in exc_message.

I might have over-indexed on the comment saying "...likes to add the query log, which we don't want to log". I took that to mean that it would be incorrect to log the query log stuff even in the debug logs (so I only logged the exc_message once that had been filtered out). If that stuff is ok to make it into debug-level logs, then I think the best approach is to log the entire exc_message (pre-split) at the top of this function.

Happy to go in whichever direction with this though!

PR #212, opened 2022-06-29
Add Grants to BigQuery Materializations

### LINKED ISSUE #198: [CT-717] Add Grants to BigQuery Materializations
blocked by: https://github.com/dbt-labs/dbt-core/issues/5263

Once the above work is merged into core, there are some tweaks that need to be made in the BigQuery adapter to fully expose the feature:
- [ ] add a call to apply_grants(..., revoke=True) to all custom materializations
- [ ] override `apply_grants` so that it directly makes the API call to revoke grants ([BigQuery Docs](https://cloud.google.com/bigquery/docs/samples/bigquery-revoke-dataset-access#bigquery_revoke_dataset_access-python)) instead of getting sql from `get_revoke_all_sql`.

emmyoop:
resolves #198

### Before merge
- [x] Reinstate original version of `dev-requirements.txt`
- [x] Ensure merging into `main` rather than a custom branch

### Description

This is dependent on https://github.com/dbt-labs/dbt-core/issues/5263

`location` is required for the sql only approach but logic needs to be added around defaulting the location.  Currently the working state has the location hard coded.

### Checklist


dbt/include/bigquery/macros/materializations/copy.sql
```diff
@@ -1,6 +1,9 @@
 {% materialization copy, adapter='bigquery' -%}
 
   {# Setup #}
+    -- grab current tables grants config for comparision later on
+  {# TODO: this is a guess I need to circle back to - does the BQ copy API also copy grants automatically?  If yes it seems like we would want to option to overwrite them if needed? #}
```
jtcohen6, line 5: > does the BQ copy API also copy grants automatically?

Great question. I don't know the answer. This materialization is also still undocumented (which is on me: https://github.com/dbt-labs/docs.getdbt.com/pull/826)

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,27 @@
...
+
+{% macro bigquery__get_revoke_sql(relation, grant_config) %}
+{# TODO: will probably need to do some more tweaks around revoke and table fields once that is resolved in core #}
+    {% for privilege in grant_config.keys() %}
+        {% set grantees = grant_config[privilege] %}
+        {% for grantee in grantees if grantee !=  target.user %}
+            revoke `{{ privilege }}` on {{ relation.type }} {{ relation.dataset }}.{{ relation.identifier }} from "{{ grantee }}";
+        {% endfor %}
+    {% endfor %}
+{% endmacro %}
```
jtcohen6, line 26: These appear to be the same as the default, except that `privilege` needs to be quoted. (I'm not sure if there's a reason why this is `{{ relation.dataset }}.{{ relation.identifier }}` instead of just `{{ relation }}`?)

I think it might make sense to add quoting to the default macros, following the adapter's `quoting` configuration and using the `adapter.quote` character. Then, we can avoid reimplementing very similar macros here as well

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,27 @@
+{% macro  bigquery__get_show_grant_sql(relation) %}
+{# Note: This only works if the location is defined in the profile.  It is an optional field right now. #}
+{# TODO: location is hardcoded for now - need logic around this (default to us)#}
+select * from {{ relation.project }}.`region-us`.INFORMATION_SCHEMA.OBJECT_PRIVILEGES
```
jtcohen6, line 4: Per [docs](https://cloud.google.com/bigquery/docs/information-schema-object-privileges):

> Queries against this view must include a [region qualifier](https://cloud.google.com/bigquery/docs/information-schema-intro#syntax).

This appears to be our first such case. We document that `location` can be set to either `US` or `EU`, but it looks like it's `None` by default. Hm.

I think we have two options:
- Pick a default if not set, probably `US` :)
- Require that folks using `grants` on BigQuery set `location` in their connection profile (not a big ask tbh)
```sql
{% set location = target.get('location', 'US') | lower %}
select * from {{ relation.project }}.`region-{{ location }}`.INFORMATION_SCHEMA.OBJECT_PRIVILEGES
```

jtcohen6 (commented):
Thanks for pushing up the draft PR !!

This also serves as a nice outline of what most adapters will need to implement to make use of the `grants` feature

dbt/include/bigquery/macros/adapters/apply_grants.sql
jtcohen6: **Update:** We've removed `{{ relation.type }}` from `default__get_grant_sql` + `default__get_revoke_sql`, because it's not required in most databases. It _is_ required in BigQuery, so I don't think we can get away from needing the overrides in this adapter.

emmyoop: @jtcohen6 `{{relation}}` works.  Can't say why I wasn't using it 😅  

As for quoting, the privilege has to be enclosed in back ticks as it's an identifier but the grantee must be enclosed in quotes (single or double) because it's a string.  Strings cannot be enclosed in backticks.

The following sql 
```sql
GRANT `roles/bigquery.dataViewer`
ON VIEW `dbt-test-env`.`emily_dataset`.`my_first_dbt_model`
TO `user:someone@email.com`;
```

throws

```
Syntax error: Expected "@" or "@@" or string literal but got identifier `user:someone@email.com` at [3:4]
```

and 

```sql
GRANT "roles/bigquery.dataViewer"
ON VIEW `dbt-test-env`.`emily_dataset`.`my_first_dbt_model`
TO "user:someone@email.com";
```

throws

```
Syntax error: Unexpected string literal "roles/bigquery.dataViewer" at [2:7]
```

It may still be worth the default macro using
```sql
grant {{ adapter.quote(privilege) }} on {{ relation.type }} {{ relation.dataset }}.{{ relation.identifier }} to {{ grantee }};
```

but this adapter will still need a custom macro for putting double/single quotes around grantee.

dbt/adapters/bigquery/impl.py
```diff
@@ -261,6 +261,29 @@ def list_relations_without_caching(
             logger.debug("list_relations_without_caching error: {}".format(str(exc)))
             return []
 
+    @available
+    def standardize_grants_dict(self, grants_table: agate.Table) -> dict:
```
jtcohen6, line 265: If this is actually the exact same as Postgres' version, should we just make this the `base` implementation?

Question also for @McKnight-42 @gshank

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,67 @@
...
+
+
+{% macro bigquery__apply_grants(relation, grant_config, should_revoke=True) %}
+
+    {% if grant_config %}
+            {% if should_revoke %}
+            {# This check is the only reason BQ needs its own copy of apply_grants - is there a better way? #}
+                {% if not target.location %}
+                    {{ exceptions.raise_compiler_error("In order to use the grants feature, you must specify a location ") }}
+                {% endif %}
```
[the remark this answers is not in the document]
emmyoop: Since `bigquery__get_show_grant_sql` uses `run_query` I wasn't sure if it was standard to throw exceptions in the macro.  Happy to move it if that's a normal pattern.

dbt/adapters/bigquery/impl.py
emmyoop: ➕

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,67 @@
+{% macro  bigquery__get_show_grant_sql(relation) %}
+    select privilege_type, grantee from {{ relation.project }}.`region-{{ target.location }}`.INFORMATION_SCHEMA.OBJECT_PRIVILEGES
+    where object_schema = "{{ relation.dataset }}" and object_name = "{{ relation.identifier }}"
```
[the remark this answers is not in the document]
emmyoop: Fair.

Is there a less yucky way than `and grantee != "user:"||session_user()`?  Just `session_user()` won't work since the grantee has `user:` prepended to it.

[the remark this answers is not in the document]
emmyoop: Of course it's not all `users`...

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,38 @@
...
+
+{% macro bigquery__get_revoke_sql(relation, grant_config) %}
+    {%- for privilege in grant_config.keys() -%}
+        {%- set grantees = [] -%}
+        {%- set all_grantees = grant_config[privilege] -%}
+        {%- for grantee in all_grantees -%}
+            {%- if grantee != target.user -%}
+                {% do grantees.append(grantee) %}
+            {%- endif -%}
+        {%- endfor -%}
```
jtcohen6, line 31: Since we're already handling the filter in the SQL (which is a better place for it):
```suggestion
        {%- set all_grantees = grant_config[privilege] -%}
```

dev-requirements.txt
```diff
@@ -1,7 +1,12 @@
 # install latest changes in dbt-core
 # TODO: how to automate switching from develop to version branches?
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-tests-adapter&subdirectory=tests/adapter
+
+git+https://github.com/dbt-labs/dbt-core.git@ct-808-grant_adapter_tests#egg=dbt-core&subdirectory=core
+git+https://github.com/dbt-labs/dbt-core.git@ct-808-grant_adapter_tests#egg=dbt-tests-adapter&subdirectory=tests/adapter
```
McKnight-42, line 5: just a heads up @emmyoop ct-808 has been merged into my [core pr](https://github.com/dbt-labs/dbt-core/pull/5369) if you want to keep these pointed at a targeted version 

```suggestion
git+https://github.com/dbt-labs/dbt-core.git@ct-660-grant-sql#egg=dbt-core&subdirectory=core
git+https://github.com/dbt-labs/dbt-core.git@ct-660-grant-sql#egg=dbt-tests-adapter&subdirectory=tests/adapter
```
should work till we merge into main.

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,29 @@
...
+            {%- endfor -%}
+        {%- endif -%}
+    {%- endfor -%}
+{%- endmacro %}
+
+
+{% macro bigquery__get_revoke_sql(relation, grant_config) %}
+    {%- for privilege in grant_config.keys() -%}
+        {%- for grantee in grant_config[privilege] -%}
+            revoke `{{ privilege }}` on {{ relation.type }} {{ relation }}  from "{{ grantee }}";
```
dbeatty10, line 26: Nit: extra space sneaked in before the `from` clause.

```suggestion
            revoke `{{ privilege }}` on {{ relation.type }} {{ relation }} from "{{ grantee }}";
```

dbeatty10: IFF BigQuery allows a comma-separated list, then you can copy the examples from Jerco [here](https://github.com/dbt-labs/dbt-core/pull/5452/files/e9899db822c72e16365fb345e8ef53bdf57d8678..16b03333f469189671c47d7127eb9bebdee77d7f) to for both `grant` and `revoke`.

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,29 @@
+{% macro  bigquery__get_show_grant_sql(relation) %}
+    {% if not target.location %}
+        {{ exceptions.raise_compiler_error("In order to use the grants feature, you must specify a location ") }}
+    {% endif %}
+
+    select privilege_type, grantee from {{ relation.project }}.`region-{{ target.location }}`.INFORMATION_SCHEMA.OBJECT_PRIVILEGES
```
dbeatty10, line 6: Learning something new everyday -- I didn't know about the information schema for [object privileges being region-specific for BigQuery](https://cloud.google.com/bigquery/docs/information-schema-object-privileges#scope_and_syntax).

An idea (which might be a bad one!):
- We could move your region-specific logic into the [BigQueryInformationSchema class](https://github.com/dbt-labs/dbt-bigquery/blob/6828a619b4ce2a88bc01770ec5e227e916c3e1d3/dbt/adapters/bigquery/relation.py#L52-L53)

That way, any user could refer to this view in a more generalized way like this and it will magically handle the region-specific bits:
`{{ relation.information_schema("object_privileges") }}`

(Not sure if I got the casing right -- maybe "object_privileges" needs to be upper-cased.)

dbt/include/bigquery/macros/materializations/incremental.sql
```diff
@@ -143,7 +143,7 @@
   {%- set tmp_relation = make_temp_relation(this) %}
 
   {#-- Validate early so we don't run SQL if the strategy is invalid --#}
-  {% set strategy = dbt_bigquery_validate_get_incremental_strategy(config) -%}
+  {%- set strategy = dbt_bigquery_validate_get_incremental_strategy(config) -%}
```
dataders, line 146:
```suggestion
  {% set strategy = dbt_bigquery_validate_get_incremental_strategy(config) -%}
```
?

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,66 @@
...
+
+{%- endmacro -%}
+
+{% macro bigquery__apply_grants(relation, grant_config, should_revoke=True) %}
+    {% if grant_config %}
+        {% if should_revoke %}
+            {% set current_grants_table = run_query(get_show_grant_sql(relation)) %}
+            {% set current_grants_dict = adapter.standardize_grants_dict(current_grants_table) %}
+            {% set needs_granting = diff_of_two_dicts_no_lower(grant_config, current_grants_dict) %}
+            {% set needs_revoking = diff_of_two_dicts_no_lower(current_grants_dict, grant_config) %}
```
emmyoop, line 49: there's a better way to do this but currently overriding this entire macro just to use a different method here that does not lower case the keys.  In Big Query the privilege is case sensitive.

test.env.example
```diff
@@ -1,3 +1,7 @@
 BIGQUERY_TEST_ALT_DATABASE=
 BIGQUERY_TEST_NO_ACCESS_DATABASE=
 BIGQUERY_TEST_SERVICE_ACCOUNT_JSON='{}'
+
+DBT_TEST_USER_1="user:buildbot@dbtlabs.com"
+DBT_TEST_USER_2="user:buildbot@fishtownanalytics.com"
```
emmyoop, line 6:
```suggestion
DBT_TEST_USER_1="group:buildbot@dbtlabs.com"
DBT_TEST_USER_2="group:buildbot@fishtownanalytics.com"
```

tests/conftest.py
```diff
@@ -30,6 +30,7 @@ def oauth_target():
         'method': 'oauth',
         'threads': 1,
         # project isn't needed if you configure a default, via 'gcloud config set project'
+        'location': 'US'
```
emmyoop, line 33: I don't think it's bets to have location as a default in the profile but I was in a crunch to try to get the tests working.  We should add location for these tests specifically and probably add a test for BQ to make sure the location is in the profile when doing grants.

tests/functional/adapter/test_grants.py
```diff
@@ -0,0 +1,44 @@
...
+        # TODO: what is insert equivalent?
+        return {"select": "roles/bigquery.dataViewer", "insert": "roles/bigquery.dataViewer", "fake_privilege": "roles/invalid"}
+
+
+    def assert_expected_grants_match_actual(self, project, relation_name, expected_grants):
+        actual_grants = self.get_grants_on_relation(project, relation_name)
+        # need a case-insensitive comparison
+        # so just a simple "assert expected == actual_grants" won't work
+        diff_a = BaseContext.diff_of_two_dicts_no_lower(actual_grants, expected_grants)
+        diff_b = BaseContext.diff_of_two_dicts_no_lower(expected_grants, actual_grants)
```
emmyoop, line 24: Over riding for case sensitivity again.

tests/functional/adapter/test_model_grants_local.py
```diff
@@ -0,0 +1,170 @@
+import pytest
```
emmyoop, line 1: This should be removed eventually.  Leaving in because when I run locally this fails in a different spot than the same test in the adapter zone.  Need more research.

emmyoop (commented):
0/5 tests pass right now.  The immediate root cause seems to be the replace logic for the privileges.  I've worked through some of it but there seems to be more.

There may also be a bug in the grant/revoke logic as the version of `test_model_grants` in this repo fails on an assertion error of grants matching.

tests/functional/adapter/test_grants.py
```diff
@@ -0,0 +1,44 @@
...
+from dbt.tests.adapter.grants.test_seed_grants import BaseSeedGrants
+from dbt.tests.adapter.grants.test_snapshot_grants import BaseSnapshotGrants
+
+from dbt.context.base import BaseContext  # diff_of_two_dicts_no_lower only
+
+
+class BaseGrantsBigQuery(BaseGrants):
+    def privilege_names(self):
+        # TODO: what is insert equivalent?
+        return {"select": "roles/bigquery.dataViewer", "insert": "roles/bigquery.dataViewer", "fake_privilege": "roles/invalid"}
```
dbeatty10, line 16: `roles/bigquery.dataEditor`?

- https://stackoverflow.com/a/65245751
- https://cloud.google.com/bigquery/docs/access-control

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -0,0 +1,66 @@
+{% macro  bigquery__get_show_grant_sql(relation) %}
+    {% if not target.location %}
+        {{ exceptions.raise_compiler_error("In order to use the grants feature, you must specify a location ") }}
+    {% endif %}
+
+    select privilege_type, grantee from {{ relation.project }}.`region-{{ target.location }}`.INFORMATION_SCHEMA.OBJECT_PRIVILEGES
```
jtcohen6, line 6: I think we need backticks around the project ID:
```suggestion
select privilege_type, grantee
from `{{ relation.project }}`.`region-{{ target.location }}`.INFORMATION_SCHEMA.OBJECT_PRIVILEGES
```

dbt/include/bigquery/macros/materializations/view.sql
```diff
@@ -9,9 +9,13 @@
 
 
 {% materialization view, adapter='bigquery' -%}
+    -- grab current tables grants config for comparision later on
+    {% set grant_config = config.get('grants') %}
+
     {% set to_return = create_or_replace_view() %}
 
     {% set target_relation = this.incorporate(type='view') %}
+    {% do apply_grants(target_relation, grant_config) %}
```
jtcohen6, line 18: We don't need this here, since the grant is already happening within `create_or_replace_view()`

test.env.example
jtcohen6: `buildbot@dbtlabs.com` + `buildbot@fishtownanalytics.com` are true aliases for one another. A grant to one is a grant to both. This causes tests to fail! We need to use genuinely different users/groups

.github/workflows/integration.yml
```diff
@@ -168,9 +168,9 @@ jobs:
           BIGQUERY_TEST_SERVICE_ACCOUNT_JSON: ${{ secrets.BIGQUERY_TEST_SERVICE_ACCOUNT_JSON }}
           BIGQUERY_TEST_ALT_DATABASE: ${{ secrets.BIGQUERY_TEST_ALT_DATABASE }}
           BIGQUERY_TEST_NO_ACCESS_DATABASE: ${{ secrets.BIGQUERY_TEST_NO_ACCESS_DATABASE }}
-          DBT_TEST_USER_1: user:buildbot@dbtlabs.com
-          DBT_TEST_USER_2: user:buildbot@fishtownanalytics.com
-          DBT_TEST_USER_3: group:dev-core@dbtlabs.com
+          DBT_TEST_USER_1: group:buildbot@dbtlabs.com
+          DBT_TEST_USER_2: group:dev-core@dbtlabs.com
+          DBT_TEST_USER_3: user:someone@dbtlabs.com # TODO
```
jtcohen6, line 173: TODO: Need to make these updates in a separate PR against `main`, so that they can be picked up in this branch (a la #216)

dbeatty10: https://github.com/dbt-labs/dbt-bigquery/pull/217 is intended to resolve this TODO.

jtcohen6: Going to change the base to `dbeatty/grantee-env-vars-take-2` and try rerunning tests!

dbeatty10 (approved):
👍 
Looks good to me. Awesome collaboration @emmyoop and @jtcohen6 to figure out all the tricky corner cases and solve them. 🙌

dbt/include/bigquery/macros/adapters/apply_grants.sql
emmyoop: This ended up being included in this PR!

McKnight-42 (approved):
This looks awesome curious if the location thing was the issue we were hitting last week?

PR #219, opened 2022-07-11
Multiple options for rendering sql for grants and revokes

dbeatty10:
### For reviewers
Could you give this description and the code a read and provide feedback of which of the options you prefer?

"Keep the implementation as-is" is Option 0 😄 

### Description

This is a refactor to give three separate things:
1. Align with base method signature: plural `grantees` rather than singular `grantee`
1. Use built-in `adapter.quote` method (rather than hard-coded backticks)
1. Use built-in `string_literal` method (rather than hard-coded double quotes)

There are three options (in addition to the original option), all of which are generalized via a helper macro:
1. Option 1: Create comma-separated values from a list 
1. Option 2: Hard-coded application of `string_literal` to a list 
1. Option 3: Application of a variable macro to a list 

### Pros of options
Using built-in methods can provide the following:
- makes implementations more similar to each other, even when there are adapter-specific customizations
- encourages standard style, which enhances readability
- encourage defaulting to usage of the built-in methods (even when they aren't strictly necessary)
- easier to see do a string search for places where quoting is being applied (without extraneous results)
- easier to copy-n-paste snippets into new adapters

### Cons of options
All of these options:
- include an additional helper method, which increases the amount of code

### Checklist

### DIFF
--- a/dbt/include/bigquery/macros/adapters/apply_grants.sql
+++ b/dbt/include/bigquery/macros/adapters/apply_grants.sql
@@ -21,6 +21,16 @@
 {%- endmacro -%}
 
 
+{%- macro bigquery__get_grant_sql__option_X(relation, privilege, grantee) -%}
+    grant {{ privilege }} on {{ relation }} to {{ grantees | join(', ') }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_revoke_sql__option_X(relation, privilege, grantee) -%}
+    revoke {{ privilege }} on {{ relation }} from {{ grantees | join(', ') }}
+{%- endmacro -%}
+
+
 {%- macro bigquery__get_grant_sql__option_0(relation, privilege, grantee) -%}
     grant `{{ privilege }}` on {{ relation.type }} {{ relation }} to {{ '\"' + grantee|join('\", \"') + '\"' }}
 {%- endmacro -%}
--- a/dbt/include/bigquery/macros/adapters/apply_grants.sql
+++ b/dbt/include/bigquery/macros/adapters/apply_grants.sql
@@ -11,10 +11,76 @@
 {% endmacro %}
 
 
-{%- macro bigquery__get_grant_sql(relation, privilege, grantee) -%}
+{%- macro bigquery__get_grant_sql(relation, privilege, grantees) -%}
+    {{ bigquery__get_grant_sql__option_3(relation, privilege, grantees) }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_revoke_sql(relation, privilege, grantees) -%}
+    {{ bigquery__get_revoke_sql__option_3(relation, privilege, grantees) }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_grant_sql__option_0(relation, privilege, grantee) -%}
     grant `{{ privilege }}` on {{ relation.type }} {{ relation }} to {{ '\"' + grantee|join('\", \"') + '\"' }}
 {%- endmacro -%}
 
-{%- macro bigquery__get_revoke_sql(relation, privilege, grantee) -%}
+
+{%- macro bigquery__get_revoke_sql__option_0(relation, privilege, grantee) -%}
     revoke `{{ privilege }}` on {{ relation.type }} {{ relation }} from {{ '\"' + grantee|join('\", \"') + '\"' }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_grant_sql__option_1(relation, privilege, grantees) -%}
+    grant {{ adapter.quote(privilege) }} on {{ relation.type }} {{ relation }} to {{ string_literal_csv(grantees) }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_revoke_sql__option_1(relation, privilege, grantees) -%}
+    revoke {{ adapter.quote(privilege) }} on {{ relation.type }} {{ relation }} from {{ string_literal_csv(grantees) }}
+{%- endmacro -%}
+
+
+{%- macro string_literal_csv(items) -%}
+    {%- for item in items -%}
+        {{ string_literal(item) }}{% if not loop.last %}, {% endif %}
+    {% endfor %}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_grant_sql__option_2(relation, privilege, grantees) -%}
+    grant {{ adapter.quote(privilege) }} on {{ relation.type }} {{ relation }} to {{ apply_string_literal(grantees) | join(', ') }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_revoke_sql__option_2(relation, privilege, grantees) -%}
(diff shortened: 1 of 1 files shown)


dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -11,10 +11,86 @@
...
+
+
+{%- macro bigquery__get_grant_sql__option_X(relation, privilege, grantee) -%}
+    grant {{ privilege }} on {{ relation }} to {{ grantees | join(', ') }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_revoke_sql__option_X(relation, privilege, grantee) -%}
+    revoke {{ privilege }} on {{ relation }} from {{ grantees | join(', ') }}
+{%- endmacro -%}
```
dbeatty10, line 31: Default implementation in dbt-core, provided for comparison purposes.

Not an option, because this syntax doesn't work for BigQuery.

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -11,10 +11,86 @@
...
 
-{%- macro bigquery__get_grant_sql(relation, privilege, grantee) -%}
+{%- macro bigquery__get_grant_sql(relation, privilege, grantees) -%}
+    {{ bigquery__get_grant_sql__option_3(relation, privilege, grantees) }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__get_revoke_sql(relation, privilege, grantees) -%}
+    {{ bigquery__get_revoke_sql__option_3(relation, privilege, grantees) }}
+{%- endmacro -%}
```
dbeatty10, line 21: Quickly swap between options by changing the macro that gets called (`*__option_{x}`).

Test via:
`python -m pytest tests/functional/adapter/test_grants.py`

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -11,10 +11,86 @@
...
+
+{%- macro bigquery__get_grant_sql__option_0(relation, privilege, grantee) -%}
     grant `{{ privilege }}` on {{ relation.type }} {{ relation }} to {{ '\"' + grantee|join('\", \"') + '\"' }}
 {%- endmacro -%}
 
-{%- macro bigquery__get_revoke_sql(relation, privilege, grantee) -%}
+
+{%- macro bigquery__get_revoke_sql__option_0(relation, privilege, grantee) -%}
     revoke `{{ privilege }}` on {{ relation.type }} {{ relation }} from {{ '\"' + grantee|join('\", \"') + '\"' }}
 {%- endmacro -%}
```
dbeatty10, line 41: Original implementation for BigQuery.

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -11,10 +11,86 @@
...
+{%- macro bigquery__get_revoke_sql__option_1(relation, privilege, grantees) -%}
+    revoke {{ adapter.quote(privilege) }} on {{ relation.type }} {{ relation }} from {{ string_literal_csv(grantees) }}
+{%- endmacro -%}
+
+
+{%- macro string_literal_csv(items) -%}
+    {%- for item in items -%}
+        {{ string_literal(item) }}{% if not loop.last %}, {% endif %}
+    {% endfor %}
+{%- endmacro -%}
```
dbeatty10, line 58: Option 1: Create comma-separated values from a list

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -11,10 +11,86 @@
...
+
+
+{%- macro apply_string_literal(items) -%}
+    {%- set apply_items = [] -%}
+    {%- for item in items -%}
+        {{ apply_items.append(string_literal(item)) }}
+    {% endfor %}
+    {%- do return(apply_items) -%}
+{%- endmacro -%}
```
dbeatty10, line 77: Option 2: Hard-coded application of `string_literal` to a list

dbt/include/bigquery/macros/adapters/apply_grants.sql
```diff
@@ -11,10 +11,86 @@
...
+
+
+{%- macro apply(macro, items) -%}
+    {%- set apply_items = [] -%}
+    {%- for item in items -%}
+        {{ apply_items.append(macro(item)) }}
+    {% endfor %}
+    {%- do return(apply_items) -%}
+{%- endmacro -%}
```
dbeatty10, line 96: Option 3: Application of a variable macro to a list.

I'm most attracted to this one because it is most generalized.

It doesn't currently accommodate multiple arguments that macro might expect though. I didn't worry about this only because the `string_literal` takes a single string argument.

It would be awesome if there were something like an `apply` [Jinja filter](https://jinja.palletsprojects.com/en/3.1.x/templates/#builtin-filters).

jtcohen6: I think like this option best, too!

> It would be awesome if there were something like an apply [Jinja filter](https://jinja.palletsprojects.com/en/3.1.x/templates/#builtin-filters).

Yes!! 

There is a `map` filter, but this can only work with other filters, not any Callable (= custom macros). Looks like this was an opinionated choice: https://github.com/pallets/jinja/issues/910

> It doesn't currently accommodate multiple arguments that macro might expect though.

How's this for a little movie magic, using [varargs](https://jinja.palletsprojects.com/en/3.0.x/templates/#:~:text=three%20special%20variables%3A-,varargs,-If%20more%20positional):

```sql
{%- macro apply(macro, items) -%}
    {%- set apply_items = [] -%}
    {%- for item in items -%}
        {{ apply_items.append(macro(item, *varargs)) }}
    {% endfor %}
    {%- do return(apply_items) -%}
{%- endmacro -%}
```

Other options:
- Use `**kwargs`
- Pass in an optional tuple of additional arguments, and then unpack them in the macro call

Example use:
```
-- macros/some_macro.sql
{%- macro append_character(string, suffix="_end") -%}
    {{ return(string ~ suffix) }}
{%- endmacro -%}
```
```
-- models/whatever.sql
{% set list_of_strings = ['string_a', 'string_b'] %}

select {{ apply(append_character, list_of_strings) }}
select {{ apply(append_character, list_of_strings, "_something_else") }}
```
```
-- compiled/whatever.sql
select ['string_a_end', 'string_b_end']
select ['string_a_something_else', 'string_b_something_else']
```

jtcohen6: If we took this approach, and developed our own generic `apply` macro, we could refactor one-offf "helper" macros that we're using today, such as [`get_quoted_csv(column_names)`](https://github.com/dbt-labs/dbt-core/blob/c25260e5dd2afa237a30db115605ece9629443d1/core/dbt/include/global_project/macros/materializations/models/incremental/column_helpers.sql#L7-L17). I think it could just become:
```
{% macro get_quoted_csv(column_names) %}
    {% set quoted_csv = apply(column_names, adapter.quote) | join(', ') %}
    {{ return(quoted_csv) }}
{% endmacro %}
```

dbeatty10:
Closing experimental PR.

PR #223, opened 2022-07-18
Changes for incremental strategy refactor compatibility

### LINKED ISSUE #232: [CT-889] Minimal changes to support dbt Core incremental materialization refactor
For compatibility, make minimal changes to support refactored code in dbt Core. In particular, making 'incremental_strategy' a config attribute means tweaking the 'config.get' call.

gshank:
resolves #232

### Description

Minimal changes to support compatibility with dbt Core incremental materialization refactor

### Checklist

### DIFF
--- a/dbt/include/bigquery/macros/materializations/incremental.sql
+++ b/dbt/include/bigquery/macros/materializations/incremental.sql
@@ -14,7 +14,7 @@
 
 {% macro dbt_bigquery_validate_get_incremental_strategy(config) %}
   {#-- Find and validate the incremental strategy #}
-  {%- set strategy = config.get("incremental_strategy", default="merge") -%}
+  {%- set strategy = config.get("incremental_strategy") or 'merge' -%}
 
   {% set invalid_strategy_msg -%}
     Invalid incremental strategy provided: {{ strategy }}
@@ -46,6 +46,12 @@
         )
       {%- endset -%}
 
+      {#-- Because we're putting the model SQL _directly_ into the MERGE statement,
+         we need to prepend the MERGE statement with the user-configured sql_header,
+         which may be needed to resolve that model SQL (e.g. referencing a variable or UDF in the header)
+         in the "dynamic" case, we save the model SQL result as a temp table first, wherein the
+         sql_header is included by the create_table_as macro.
+      #}
       {{ get_insert_overwrite_merge_sql(target_relation, source_sql, dest_columns, [predicate], include_sql_header=true) }}
 
   {% else %} {# dynamic #}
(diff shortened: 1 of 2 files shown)


dbt/include/bigquery/macros/materializations/incremental.sql
```diff
@@ -46,7 +46,9 @@
         )
       {%- endset -%}
 
-      {{ get_insert_overwrite_merge_sql(target_relation, source_sql, dest_columns, [predicate], include_sql_header=true) }}
+      {{ config.set('include_sql_header', true) }}
+      {{ get_insert_overwrite_merge_sql(target_relation, source_sql, dest_columns, [predicate]) }}
+      {{ config.set('include_sql_header', false) }}
```
jtcohen6, line 51: Oh, this is clever! And does not feel any hackier than what we're doing right now.

Could we add a comment here, to the effect of:
```sql
/* {#
    because we're putting the model SQL _directly_ into the MERGE statement,
    we need to prepend the MERGE statement with the user-configured sql_header,
    which may be needed to resolve that model SQL (e.g. referencing a variable or UDF in the header)
    in the "dynamic" case, we save the model SQL result as a temp table first, wherein the
    sql_header is included by the create_table_as macro
#} */
```

jtcohen6 (commented):
I'm glad we didn't push the refactor all the way through in every repo, so we can see the (non-)impact on adapters that (a) have their own `incremental` materialization, and (b) haven't yet implemented the required pieces of the new strategy mechanism.

jtcohen6:
@gshank It does look a test started failing, relevant to the `include_sql_header` functionality:
```
FAILED tests/integration/bigquery_test/test_simple_bigquery_view.py::TestUnderscoreBigQueryRun::test_bigquery_run_twice
```
```
00:54:57  Database Error in model sql_header_model_incr_insert_overwrite_static (models/sql_header_model_incr_insert_overwrite_static.sql)
00:54:57    Function not found: a_to_b at [21:5]
00:54:57    compiled SQL at target/run/test/models/sql_header_model_incr_insert_overwrite_static.sql
```
I'm very glad we have a test for it, in some way shape or form

dbt/include/bigquery/macros/materializations/incremental.sql
```diff
@@ -46,6 +46,12 @@
         )
       {%- endset -%}
 
+      {#-- Because we're putting the model SQL _directly_ into the MERGE statement,
```
nathaniel-may, line 49: do we need comment symbols on each line?

nathaniel-may (approved):

dbt/include/bigquery/macros/materializations/incremental.sql
McKnight-42: jinja does support comment blocks if thats what your referring to https://tedboy.github.io/jinja2/templ5.html, https://github.com/HubSpot/jinjava/blob/master/src/test/resources/parse/tokenizer/multiline-comment.jinja.

McKnight-42 (approved):

dbt/include/bigquery/macros/materializations/incremental.sql
gshank: The reason to put in the SQL comment lines is not because they help comment but because it makes the file highlighting look better :)

github-christophe-oudar: An open question regarding that `sql_header`: is there a plan to detect them automatically these kind of headers to avoid the "unfriendly" approach of wrapping the code in a `{%- call set_sql_header(config) %}` block.
It requires dedicated documentation and mentoring on our end and I feel it could be simplified but maybe it's too much work for the overall usage?

PR #225, opened 2022-07-19
Add job_id/project_id to adapter response to enable easy job linking

### LINKED ISSUE #92: Provide jobIds behind model creation in run_results.json
In BigQuery, each time SQL is executed, a job is created and assigned a unique ID. _(Example: `job_9JVepH9O1bzemNfB6xZzxz_mfFkY`)_ When DBT runs it checks for the results/status of these jobs.

We've found the `run_results.json` file extremely useful, and being able to tie these results (errors, compiled SQL, variables, tags, etc.) back to the specific job that executed in BigQuery for additional information and troubleshooting would make `run_results.json` even more useful.

You could try to look at a combination of DBT logs and BigQuery logs in Stackdriver Logging and try to approximate which exact job corresponded to the model creation by timestamp proximity or query (compiled SQL), but those methods are not deterministic.

### Additional context
This is specific to BigQuery for now, but I am sure there are parallels in other databases (transaction ID? query ID?).

Another aspect to consider would be whether to provide multiple `jobId`s if there is a macro called in the model that executes additional queries against the database.

### Who will this benefit?
Anyone that wants to tie run results back to specific jobs for additional debugging would benefit from this feature.

Kayrnt:
resolves #92 

### Description

Adding job_id / project_id from the job executed by BQ to add a clean link to explore the runtime in the BQ UI.

Practically, it adds a link in your `-d` runs such as:
```
13:28:11.169553 [debug] [Thread-1 (]: BigQuery adapter: Job details accessible at https://console.cloud.google.com/bigquery?project=<project>&j=<jobId>&page=queryresults
```

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -86,6 +86,9 @@ class BigQueryConnectionMethod(StrEnum):
 @dataclass
 class BigQueryAdapterResponse(AdapterResponse):
     bytes_processed: Optional[int] = None
+    location: Optional[str] = None
+    project_id: Optional[str] = None
+    job_id: Optional[str] = None
 
 
 @dataclass
@@ -188,6 +191,12 @@ class BigQueryConnectionManager(BaseConnectionManager):
     @classmethod
     def handle_error(cls, error, message):
         error_msg = "\n".join([item["message"] for item in error.errors])
+        if hasattr(error, "query_job"):
+            logger.error(
+                cls._bq_job_link(
+                    error.query_job.location, error.query_job.project_id, error.query_job.job_id
+                )
+            )
         raise DatabaseException(error_msg)
 
     def clear_transaction(self):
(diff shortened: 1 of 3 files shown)


dbt/adapters/bigquery/connections.py
```diff
@@ -437,51 +439,67 @@ def execute(
...
+        job_id = None
+        project_id = None
+        num_rows_formatted = None
+        processed_bytes = None
+
+        def set_common_attributes():
+            nonlocal message, num_rows, num_rows_formatted, bytes_processed, processed_bytes, job_id, project_id
+            bytes_processed = query_job.total_bytes_processed
+            processed_bytes = self.format_bytes(bytes_processed)
+            job_id = query_job.job_id
```
MartinNowak, line 451: At least for the tabbed BigQuery Console UI in in EU, I need a qualified job id.
```suggestion
            job_id = f"bq:{query_job.location}:{query_job.job_id}"
```

github-christophe-oudar: Thanks for the feedback, as I'm running in US region, I didn't know about it.
Yet I would put the location in a field and add the prefixes in the logged url as some people might use the returned value in adapter response.

MartinNowak:
It's a bit hidden in the massive debug logs.
It also doesn't log the job link for failing queries (syntax errors raised via handle_error).
Furthermore logging the link separately leads to interleaved output with multiple threads.

Here are WIP patches we're using atm.
Immediately clickable query debugging is a major productivity boost.
https://github.com/7Learnings/dbt-core/tree/v1.2.0+bigquery-job-links.1 (add job link to adapter response message)
https://github.com/7Learnings/dbt-bigquery/tree/v1.2.0+bigquery-job-links.1 (log adapter response for failing tests)

I'd love to properly integrate this with dbt, but never managed to find the sweet spot in the logging system.

github-christophe-oudar:
I agree it's not perfect, I hesitated to add it to the actual "message" because I preferred to have a dedicated log.

If we have to modify the core as well, maybe @jtcohen6 would like to be involved to be sure so that we can have some feedback regarding providing a smart interface for other adapters to provide a custom log (that would make sense to me).

McKnight-42:
@Kayrnt thank you so much for opening this PR! While testing this locally I noticed we aren't applying the `set_common_attributes` method as we never call it outside of itself. so we aren't currently seeing the `job_id` or `project_id` in the `adapter_response` ex.` adapter_response={'_message': 'OK', 'code': 'CREATE VIEW'}` which is causing a `KeyError` failure on `test__bigquery_adapter_functions`.

McKnight-42:
@Kayrnt thank you so much for opening this PR! While testing this locally I noticed we aren't applying the `set_common_attributes` method as we never call it outside of itself. so we aren't currently seeing the `job_id` or `project_id` in the `adapter_response` ex.` adapter_response={'_message': 'OK', 'code': 'CREATE VIEW'}` which is causing a `KeyError` failure on `test__bigquery_adapter_functions`.  Do you have a sense of where it is your wanting to call upon it in the execute method? @jtcohen6 any thoughts?

dbt/adapters/bigquery/connections.py
github-christophe-oudar: I've added the location in the response & log 👍

github-christophe-oudar:
@McKnight-42 Thanks for the feedback, I used `set_common_attributes` within the `execute` function in all blocks. I didn't use it for `CREATE_VIEW` because it wasn't putting a message but I guess I could use `set_common_attributes` as well so I'm updating to be consistent.
I guess that if I'm doing it for all query types, then it's just clearer to inline it at the end of the specific parts.

tests/integration/bigquery_test/test_bigquery_adapter_functions.py
```diff
@@ -21,6 +21,12 @@ def test__bigquery_adapter_functions(self):
         results = self.run_dbt()
         self.assertEqual(len(results), 3)
 
+        for result in results:
+            # all queries in adapter models are jobs that are expected to have a job_id
+            assert result.adapter_response["location"] is not None
+            assert result.adapter_response["job_id"] is not None
+            assert result.adapter_response["project_id"] is not None
```
jtcohen6, line 28: 👍 nice test!

dbt/adapters/bigquery/connections.py
```diff
@@ -446,55 +455,70 @@ def execute(
...
+            location=location,
+            project_id=project_id,
+            job_id=job_id,
         )
 
         return response, table
 
+    @staticmethod
+    def _bq_job_link(location, project_id, job_id) -> str:
+        return f"https://console.cloud.google.com/bigquery?project={project_id}&j=bq:{location}:{job_id}&page=queryresults"
```
jtcohen6, line 520: This is neat! Do you think there's any risk of this URL format changing in the next few years?

github-christophe-oudar: We have that logic in a inhouse too for more than 3 years, hopefully it will stays the same for another 3 years? Anyway if it changes it should be a painless update though indeed it's not 100% futureproof

McKnight-42:
@Kayrnt appears we may of had a slight false positive here after the changes due to a `okay to test` flag missing not triggering the `AdapterIntegrationTests`.  It looks like we are still getting the `KeyError` issue as seen here [ci/cd tests on main](https://github.com/dbt-labs/dbt-bigquery/runs/7677150275?check_suite_focus=true#step:7:629). Please feel free to reopen this pr if you have time to keep working on this issue/refactor.

github-christophe-oudar:
@McKnight-42 OK I'll have a look and try to reproduce it locally 👍

dbt/adapters/bigquery/connections.py
MartinNowak: They even preserved it across their full tabbed UI rewrite (which also changed every other URL format).

PR #226, opened 2022-07-20
Add incremental materializations

stu-k:
### Description
Implements the incremental materialization for bigquery. Nearly directly copied from [Ian's work in spark](https://github.com/dbt-labs/dbt-spark/pull/377/files#diff-786bb6587e86e50a2d01888eb2d4a5257e9a0025f75379214fa93cd5a033c9fbR131)

### DIFF
--- a/dbt/include/bigquery/macros/materializations/incremental.sql
+++ b/dbt/include/bigquery/macros/materializations/incremental.sql
@@ -221,13 +221,6 @@
     {% endcall %}
 
     {%- if language == 'python' and tmp_relation -%}
-      {#--
-      This is yucky.
-      See note in dbt-spark/dbt/include/spark/macros/adapters.sql
-      re: python models and temporary views.
-
-      Also, why does not either drop_relation or adapter.drop_relation work here?!
-      --#}
       {{ adapter.drop_relation(tmp_relation) }}
     {%- endif -%}
 
--- a/dbt/include/bigquery/macros/materializations/incremental.sql
+++ b/dbt/include/bigquery/macros/materializations/incremental.sql
@@ -1,13 +1,14 @@
-{% macro declare_dbt_max_partition(relation, partition_by, sql) %}
+{% macro declare_dbt_max_partition(relation, partition_by, complied_code, language='sql') %}
 
-  {% if '_dbt_max_partition' in sql %}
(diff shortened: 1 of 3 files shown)


dbt/adapters/bigquery/impl.py
```diff
@@ -215,6 +215,33 @@ def check_schema_exists(self, database: str, schema: str) -> bool:
             return False
         return True
 
+    @available.parse_none
+    def submit_python_job(self, parsed_model: dict, compiled_code: str):
```
jtcohen6, line 219: I don't think we need this here

jtcohen6 (commented):
@stu-k Thanks for pushing up the initial PR!

As discussed during our chat yesterday, there are (I believe) two cases in which we do not create a "temp" table before merging:
- `merge` strategy with `on_schema_change: ignore`
- `insert_overwrite` strategy with "static" `partitions` config

In those cases, dbt templates the model `{{ sql }}` directly into the `merge` statement. That won't work for Python, so we really do need to submit the Python job and save the results to a table before the `merge` takes place. We also need to make sure that table is cleaned up after the `merge`, either by altering it to set it to expire within X hours (our standard approach is 12), or drop it explicitly at the end of the materialization.

stu-k:
@jtcohen6 Apologies I didn't mean to add you as a reviewer, this is still in progress, I'm just using the pr to run the CI.

jtcohen6:
@stu-k Oh whoops, sorry! Keep on keepin' on!

dbt/include/bigquery/macros/materializations/incremental.sql
```diff
@@ -160,43 +164,74 @@
...
+    {%- if language == 'python' and tmp_relation -%}
+      {#--
+      This is yucky.
+      See note in dbt-spark/dbt/include/spark/macros/adapters.sql
+      re: python models and temporary views.
 
-  {%- call statement('main') -%}
-    {{ build_sql }}
-  {% endcall %}
+      Also, why does not either drop_relation or adapter.drop_relation work here?!
```
ChenyuLInx, line 229: Should we remove this comment?

ChenyuLInx (approved):
Looks great to me!

PR #230, opened 2022-07-20
retry transient connection errors

### LINKED ISSUE #229: [CT-878] bigquery: retry transient connection exceptions
Google has a built in way to do this, so we may want to use that instead of conforming to the way other dbt adapters manage retries via the core helper `retry_connection`.

nathaniel-may:
resolves #229 

Instead of matching the other dbt adapters, this pr instead leverages the provided retry decorator from Google to retry with their default settings.

This doesn't give users the option to configure anything with regards to connection retries. Only retrying transient errors as defined by bigquery library maintainers seems like solid enough behavior to not need overrides but I'd like to hear if anyone feels differently.

### Checklist


dbt/adapters/bigquery/connections.py
```diff
@@ -305,6 +305,7 @@ def get_impersonated_bigquery_credentials(cls, profile_credentials):
         )
 
     @classmethod
+    @retry.Retry()  # google decorator. retries on transient errors with exponential backoff
```
jtcohen6, line 308: _Live discussion_

This is Google's own `Retry()` context manager. It has really sensible defaults and behavior (much better than ours!), and we don't think there's any risk in just putting it everywhere. The scope of the current work is standardizing how we open connections, though, and this is the place to do that.

Follow-up work:
- Add this `Retry()` decorator in more places (specifically, in place of our current homespun approach via `_retry_and_handle`)
- If the user provides configuration for the `job_*` retry/timeout options, plumb them into the `Retry()` decorator in the appropriate ways. It is reasonable that a user would want a very quick timeout if a job fails to create/submit, and a much much _much_ longer timeout for a job that's actually running (= models that take >1 hr to run).

nathaniel-may: this is captured in #231 now

jtcohen6 (approved):
Looks good! This might even help our failures in CI, where we've lately been hitting more "service is currently unavailable" errors.

Let's look to schedule the larger refactoring work in https://github.com/dbt-labs/dbt-bigquery/issues/231 for an upcoming sprint, or later this year.

dbt/adapters/bigquery/connections.py
hui-zheng: reposting from [#231-comment](https://github.com/dbt-labs/dbt-bigquery/issues/231#issuecomment-1231026219)

I agree/like this approach. it's much cleaner and leverage the more robust Google Retry mechanism.

One thing to call out.
Dbt currently has two ways to config bq job retry: `job_retry_deadline_seconds` and `job_retries`. 

However, [Google Retry](https://googleapis.dev/python/google-api-core/latest/retry.html#google.api_core.retry.Retry) only uses `job_retry_deadline_seconds`, to control retries, not `job_retries`. 

when moving over to Google Retry, I suggest 
 - removing `job_retries` config
 - expand `job_retry_deadline_seconds` into two configs: `job_creation_retry_deadline_seconds` and `job_execution_retry_deadline_seconds`
this provide the proper control over retries in both `job_creation` and `job_execution` steps.

colin-rogers-dbt:
@nathaniel-may / @jtcohen6 any reason this wasn't merged?

nathaniel-may:
No. I was transitioning to EM and it fell off my radar.

PR #259, opened 2022-08-07
Experiment with Dataproc Serverless

### LINKED ISSUE #248: [CT-980] [feature] Use dataproc serverless instead of dataproc cluster
# Description
> This issue is updated from a spike to a normal issue, original description below

Update python model submission methods in dataproc to support both Dataproc cluster and Dataproc serverless. When a cluster ID is provided, we are going to use the dataproc cluster by default. User can overwrite it by specify a separate config in the yml file of yml file. When cluster ID is not provided, we will submit using serverless by default, user can overwrite it by provide a cluster ID and submission method in the config for model.

- [ ] being able to run with cluster and serverless at the same time
- [ ] enable tests for python again(running with serverless)

## Original description
Good context in [this Slack thread](https://getdbt.slack.com/archives/C03QUA7DWCW/p1659470376608489)

Docs: https://cloud.google.com/dataproc-serverless/docs/overview

Dataproc Serverless:
- does not require an always-on cluster
- does not require end users to provide a `dataproc_cluster_name` + `dataproc_region` in their connection profile
- may have better handling for concurrency / auto-scaling

Risks to investigate:
- [ ] Less straightforward to set up initially — we would need to clearly document any pitfalls
- [ ] Potentially more costly per vCPU hour (but much more cost-effective overall)
- [ ] Longer runtimes, because dbt would have to submit "batch" PySpark jobs
- [ ] Installing third-party packages is less straightforward, since they cannot be configured on the cluster at creation time. The mechanisms here seem to be a custom image + `conda-pack` ([docs](https://conda.github.io/conda-pack/index.html)). The default image for Dataproc Serverless ([docs](https://cloud.google.com/dataproc-serverless/docs/guides/custom-containers#dockerfile)) seems to include a few popular packages (`koalas`, `numpy`, `pandas`, `regex`, `scikit-learn`, ...), but this is far from everything a user would need.

jtcohen6:
resolves #248

### Description

**What's cool?** No need to provision a Dataproc Cluster in advance! No need to pay to keep it on 24/7! Sensible defaults around cluster sizing, and auto-scaling to handle concurrency.

The only required setup is enabling Dataproc APIs for a region, and some networking setup to enable "Private Google Access."

<details><summary>Screenshot</summary>
Screenshot 2022-08-07 at 17 32 07
</details>

**What's less cool?**
- **SLOW.** The Dataproc UI very clearly states that "Spark jobs take ~60 seconds to initialize resources." Additionally, because Dataproc Serverless submits batch jobs asyncronously, I noticed an additional delay between the job finishing, and dbt figuring that out. All in, the dbt model took 223.19s (3+ minutes) to run, the Dataproc elapsed time was 1 min 22 sec, and the "actual" run time was 34s.
- **Third-party packages:** Using the default container, which only contains a small number of desirable packages. No really good way to add more, short of having dbt try to run Docker for you—I strongly feel that we should avoid doing this!! We could enable users to spin up their own custom containers, and pass the pointer to that container as a model configuration. Also, it is possible to pass the GCS locations of additional Python files, which could be good for first-party packages.

Maybe a dedicated Dataproc cluster in development, and Dataproc Serverless in production...?

### Checklist

- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

### DIFF
--- a/dbt/adapters/bigquery/impl.py
+++ b/dbt/adapters/bigquery/impl.py
@@ -898,7 +898,7 @@ class DataProcHelper:
         self.storage_client = storage.Client(
             project=self.credential.database, credentials=self.GoogleCredentials
         )
-        self.job_client = dataproc_v1.JobControllerClient(
+        self.job_client = dataproc_v1.BatchControllerClient(
             client_options={
                 "api_endpoint": "{}-dataproc.googleapis.com:443".format(
                     self.credential.dataproc_region
@@ -913,20 +913,34 @@ class DataProcHelper:
         blob.upload_from_string(compiled_code)
 
     def submit_dataproc_job(self, filename: str):
-        # Create the job config.
-        job = {
-            "placement": {"cluster_name": self.credential.dataproc_cluster_name},
-            "pyspark_job": {
-                "main_python_file_uri": "gs://{}/{}".format(self.credential.gcs_bucket, filename)
-            },
+        from google.cloud import dataproc_v1
+
+        # create the Dataproc Serverless job config
+        batch = dataproc_v1.Batch()
+        gcs_location = "gs://{}/{}".format(self.credential.gcs_bucket, filename)
+        batch.pyspark_batch.main_python_file_uri = gcs_location
+        batch.pyspark_batch.jar_file_uris = [
+            "gs://spark-lib/bigquery/spark-3.1-bigquery-0.26.0-preview.jar"  # how to keep this up to date?
+        ]
+
+        # should we make all of these spark/dataproc properties configurable?
+        # https://cloud.google.com/dataproc-serverless/docs/concepts/properties
+        # https://cloud.google.com/dataproc-serverless/docs/reference/rest/v1/projects.locations.batches#runtimeconfig
+        batch.runtime_config.properties = {
+            "spark.executor.instances": "2",
         }
-        operation = self.job_client.submit_job_as_operation(
-            request={
-                "project_id": self.credential.database,
-                "region": self.credential.dataproc_region,
-                "job": job,
-            }
+
+        parent = f"projects/{self.credential.database}/locations/{self.credential.dataproc_region}"
+
+        request = dataproc_v1.CreateBatchRequest(
+            parent=parent,
+            batch=batch,
         )
+
+        # make the request
+        operation = self.job_client.create_batch(request=request)
+        # this takes quite a while
+        # https://googleapis.dev/python/google-api-core/latest/operation.html#google.api_core.operation.Operation
         response = operation.result()
         return response
 


velascoluis:
Hi, great to see spark serverless support on dbt!
A few of comments If I may
1 - Do you think the network should be also configurable at the profile level? As per [this](https://cloud.google.com/dataproc-serverless/docs/concepts/network), it is very common to use a specific network for prod. workloads

```
batch.pyspark_batch.main_python_file_uri = gcs_location
        batch.pyspark_batch.jar_file_uris = [
            "gs://spark-lib/bigquery/spark-3.1-bigquery-0.26.0-preview.jar"  # how to keep this up to date?
        ]

      
        batch.runtime_config.properties = {
            "spark.executor.instances": "2",
        }

batch.environmentConfig.executionConfig = {
                    "subnetworkUri": "{}".format(self.credential.dataproc_subnet)
                }
```

Also, as per [this](https://cloud.google.com/dataproc/docs/tutorials/bigquery-connector-spark-example), perhaps it's a good idea to use the latest releases of the connector? (havent tried myself)

```
velascoluis-macbookpro:dbt-bigquery velascoluis$ gsutil ls gs://spark-lib/bigquery/spark*latest*

gs://spark-lib/bigquery/spark-2.4-bigquery-latest.jar
gs://spark-lib/bigquery/spark-bigquery-latest.jar
gs://spark-lib/bigquery/spark-bigquery-latest_2.11.jar
gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar
```

3- Is is possible to add a timeout for the operation as we have with BQ?

Thanks

jtcohen6:
@velascoluis Thanks for catching this so quickly! This is definitely still "experimental" code while we're investigating. We're relatively new to this tool set, so very very happy to have advice from actual practitioners at GCP :)

Rather than add many many additional configurations to `profiles.yml`, I'm thinking that we'd probably want to make each of these configurations available per model. That would also take advantage of another value prop of Dataproc Serverless — each model is its own independent "batch job," and can declare the properties of the infrastructure it wants to run on. We'll hard-code some things (always PySpark file URI, always use latest spark-bigquery-connector JAR), and leave the rest up to the user:
```yml
version: 2

models:
  - name: my_python_model
    config:
      pyspark_batch:
        args:
          - ...
      runtime_config:
        spark.executor.instances: 2
      execution_config:
        subnetwork_uri: ...
```

Similar logic could apply for "traditional" Dataproc (managed clusters), too — we could have a default cluster configured in `profiles.yml`, but then also give users the ability to override the cluster used for one particular model. It seems like we may want to enable both patterns, given the relative advantages of each.

Re: timeout/retry: We'd like to start using Google's built-in retry capabilities as much as possible (#230), rather than rolling this logic ourselves. Could this be as simple as adding the `Retry()` decorator where we're submitting the Dataproc job/batch?

velascoluis:
> Rather than add many many additional configurations to `profiles.yml`, I'm thinking that we'd probably want to make each of these configurations available per model. 

This makes a lot of sense, and will allow - as you mentioned before - to use dataproc serverless for workloads such as development / iterative or exceptional heavy one-shots (e.g. generating job templates with opinionated defaults) , leaving  standard dataproc for daily/stable ELTs where cost/config is more predictable. 

> Re: timeout/retry: We'd like to start using Google's built-in retry capabilities as much as possible (#230), rather than rolling this logic ourselves. Could this be as simple as adding the `Retry()` decorator where we're submitting the Dataproc job/batch?

Yes, I was thinking on crafting spark properties to the job itself and control specific timeouts (network, idle workers ..), but delegating the responsibility seems more sensible and general approach.

Btw, as Im aware of some deployments of dbt core on GKE, I did some tests with dataproc on GKE, and it works fine as well (same API)

jtcohen6:
Closed in favor of https://github.com/dbt-labs/dbt-bigquery/pull/303

Issue #17, opened 2020-03-31
allow `_dbt_max_partition` to be disabled


hui-zheng:
### Describe the feature

In dbt v0.16, dbt run will make query calls for  `_dbt_max_partition` on all BigQuery partitioned incremental models. 

In our situation, we have many very large partitioned incremental models that have no use of `_dbt_max_partition`. However, they still have to query  `_dbt_max_partition`. It is very costly when the runs in production add up.

Could we disable `_dbt_max_partition` for models that have no use of it?

### Describe alternatives you've considered

To allow the model config to disable   `_dbt_max_partition` for a particular model. 

for example,
```
{{ config(
    materialized='table',
    partition_by={
      "field": "created_at",
      "data_type": "timestamp",
      "enable_dbt_max_partition": false
    }
)}}
```

### Additional context

This is specific to BigQuery database, and it's part of the new merge feature from dbt v0.16

### Who will this benefit?

Everyone who own very large BigQuery tables for incremental runs but not using   `_dbt_max_partition`

PR #285, opened 2022-08-23
Add macro which returns the largest partition

### LINKED ISSUE #286: [CT-1071] Macro which returns the largest partition of a table
### Describe the feature

A macro which uses the `INFORMATION_SCHEMA` to return a literal defining the maximum partition of an incremental model to avoid having to query `{{ this }}` in an `if is_incremental()` block.

### Who will this benefit?
People who have large incremental models.

Yup.

judahrand:
resolves #286 

### Description

This macro uses the `INFORMATION_SCHEMA` to cheaply obtain the maximum partition of a BigQuery table in order to reduce the cost of an incremental model. For example:

```sql
SELECT *
FROM source('my', 'source')
{% if is_incremental() %}
WHERE timestamp >= {{ get_max_partition(this) }}
{% endif %}
```

The macro is billed at 0 bytes whereas using the typical `SELECT MAX(DATE(timestamp)) FROM {{ this }}` would be billed and could be extremely expensive if the model is very large.

### Checklist


dbeatty10:
Nice @judahrand 🤩 

My understanding is that there are two main options for getting the maximum timestamp from BigQuery tables for the purposes of incremental materializations:
1. Current implementation using `select max({{ partition_by.field }})` (which uses Google Standard SQL)
1. Alternative implementation using [`adapter.get_partitions_metadata`](https://github.com/dbt-labs/dbt-bigquery/blob/ad885cd6db8b3a26278cb16d452038b061aa88e8/dbt/adapters/bigquery/connections.py#L526-L535) (which uses `$__PARTITIONS_SUMMARY__` in legacy SQL)

You make a compelling case to update the implementation of the incremental materialization for bigquery to use the less expensive one (second option):
> The macro is billed at 0 bytes whereas using the typical SELECT MAX(DATE(timestamp)) FROM {{ this }} would be billed and could be extremely expensive if the model is very large.

However, it looks like the default implementation was _intentionally_ left as expensive and users are that empowered to override the `declare_dbt_max_partition()` macro in their project with their own implementation (like the one you included in this PR).

Here's a lightly edited selection from #41 where I'm pulling this intent from:

> - Rather than hard-coding the declaration of `_dbt_max_partition` in the dynamic insert-overwrite script, shell out to a macro `declare_dbt_max_partition()` that users can override with custom logic
> - Keep the default logic simple (+ expensive) for now
> - That custom logic could include:
>     - disabling this variable entirely
>     - using `get_partitions_metadata` (i.e. querying `$__PARTITIONS_SUMMARY__` with legacySQL)
>     - using `INFORMATION_SCHEMA.PARTITIONS` (standardSQL)

https://github.com/dbt-labs/dbt-bigquery/issues/17 also mentions this idea:
> Perhaps we should add support for a new config that would override the default value used within dbt today. So, for instance:

> If the user wishes, they could set the config using adapter.get_partitions_metadata(target_relation) to get a $0 dynamic metadata lookup (though using legacySQL). We could offer a special config value or recommended code snippet for this.

@jtcohen6 is this an accurate understanding that we plan to keep the default implementation of `declare_dbt_max_partition` to use Google Standard SQL? Or is the option on the table to update the default implementation to one that uses BigQuery legacy SQL? Or is there another implementation option for saving cost that we'd be willing to consider?

judahrand:
@dbeatty10 Sorry I've let this sit for so long...

While the original implementation may have been deliberately left as 'expensive' I don't see that as a justification for continuing to make it expensive...

As far as I can see there's no real downside to this new implementation other than a small amount of complexity? It should still be possible to override the `declare_dbt_max_partition` and afaik the new implementation doesn't need any additional permissions.

willbryant:
I'm with @judahrand.

I don't ready the quoted text above to mean that it was a *feature* to be using an expensive query by default, just that that was the status quo, and that that could be made configurable.

I don't personally see how anyone would be disadvantaged by using a better implementation by default, but there is a behavior change here; the current implementation isn't really finding the latest partition, it's finding the latest partition *with rows in it*. This seems like a bug to me given the name of the call, but thought I'd flag it.

NB. I am particularly +1 on this PR because it also fixes the problem of not being able to access the latest partition found for a different output table; I want that because I want to be able to easily make a "all the data since the materialized stuff" view.

Perhaps if there's no clear outcome on this MR yet, the refactor to a macro could be done in a separate PR to the change to the implementation, so they can be discussed separately?

Fleid:
@dbeatty10 I think your concern is valid.

Here, since there's no movement towards a more modern take (it looks like the [potential "modern" alternative](https://github.com/dbt-labs/dbt-core/issues/3666#issuecomment-929526823) is still in [preview](https://cloud.google.com/bigquery/docs/information-schema-partitions)), I'm not against going "backward".

I'll flag this for code review and line it up in the backlog.
We will need to add tests though.

dbeatty10:
Sounds awesome @Fleid -- thanks for giving this a look and moving it forward!

willbryant:
In November, I mentioned:

> there is a behavior change here; the current implementation isn't really finding the latest partition, it's finding the latest partition with rows in it. This seems like a bug to me given the name of the call, but thought I'd flag it.

I noticed today a problem with a model I have added that sometimes has no new rows to add (on incremental runs) - the current merge SQL creates a `__NULL__` partition when there's no rows produced by the incremental source query. This is odd since there's no rows in the new partition.

I am surprised by this behavior, but with the current _dbt_max_partition behavior, we get away with it, because partitions with no rows aren't returned.

However, after this MR, the `| max` unfortunately returns the `__NULL__`, even though there's no rows in that partition.

judahrand:
> In November, I mentioned:
> 
> > there is a behavior change here; the current implementation isn't really finding the latest partition, it's finding the latest partition with rows in it. This seems like a bug to me given the name of the call, but thought I'd flag it.
> 
> I noticed today a problem with a model I have added that sometimes has no new rows to add (on incremental runs) - the current merge SQL creates a `__NULL__` partition when there's no rows produced by the incremental source query. This is odd since there's no rows in the new partition.
> 
> I am surprised by this behavior, but with the current `_dbt_max_partition` behavior, we get away with it, because partitions with no rows aren't returned.
> 
> However, after this MR, the `| max` unfortunately returns the `__NULL__`, even though there's no rows in that partition.

It shouldn't be hard to filter out partitions with no rows in them I don't think. Or alternatively just filter out the `__NULL__` partition entirely?

judahrand:
> @dbeatty10 I think your concern is valid.
> 
> Here, since there's no movement towards a fresher take (it looks like the [potential "modern" alternative](https://github.com/dbt-labs/dbt-core/issues/3666#issuecomment-929526823) is still in [preview](https://cloud.google.com/bigquery/docs/information-schema-partitions)), I'm not against going "backward".
> 
> I'll flag this for code review and line it up in the backlog. We will need to add tests though.

I do actually think that using the 'new' INFORMATION_SCHEMA is probably the way to go even if it is 'preview'. It will allow us to filter on the number of rows in the partition which should help avoid edge cases where partitions exist but are empty.

nssalian:
@judahrand, thanks for the work here. You should add a functional test to make sure we get the expected value in the different formats. Also, a branch update might be needed.

richardbiddle:
Please consider the case where partition_id can be '__STREAMING_UNPARTITIONED__', maybe need to exclude like the __NULL__ case?

mwstanleyft:
There's another spot [here](https://github.com/dbt-labs/dbt-bigquery/blob/b06d2309cff249f74b8aea02d0c56e8d763f943c/dbt/include/bigquery/macros/materializations/incremental_strategy/insert_overwrite.sql#L151-L155) where the adapter is going to scan the whole partition column as well. It would be a good idea I think to pull this out into its own macro so it can be shimmed, and also offer the option to use INFORMATION_SCHEMA instead.

Accessing INFORMATION_SCHEMA requires these permissions, and I'm not sure all of these are required at present. So using INFORMATION_SCHEMA may break people's jobs where their service account doesn't have enough permission.

```
bigquery.tables.get
bigquery.tables.getData
bigquery.tables.list
```

https://cloud.google.com/bigquery/docs/information-schema-partitions

judahrand:
> Any updates here? This would be great to get merged.

I'd happily deal with the merge conflicts if someone would assure me that it would in fact be merged 🤷‍♂️

bmiselis:
Up, it'd be great to see this being merged.

Chr96er:
This "needs" to be merged. The current way of retrieving the maximum partition is extremely and unnecessarily expensive for large tables. I haven't seen any pushback except for requests to implement tests.

I've used the macros myself but some of my tables have `__NULL__` partitions and I don't know how to apply the `get_partitions_metadata` that is proposed in this PR (if it's possible at all without running a fork of DBT).

PR #238, opened 2022-07-26
Using information schema

hassan-mention-me:
resolves #113 https://github.com/dbt-labs/dbt-bigquery/issues/113

Using `information_schema` instead of __TABLES__ to fetch catalog data, which limits access level needed to generate DBT docs.
### Checklist

- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-bigquery next" section.

Issue #260, opened 2022-08-08
[CT-1009] [CT-1008] [Bug] Catalog query regex is too aggressive at detecting sharded tables


adamcunnington-mlg:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

The following applies to BigQuery (not tested with other adapters):

If you define a source table where the underlying relation is a view, dbt cloud successfully "links" it to that relation and shows complete information; details, columns, sample SQL etc. Fantastic!

However, if the source table's underlying relation a table, it does not link. dbt Cloud docs do not show the relation nor (presumably therefore) the columns or a column list in the sample sql.

The ONLY thing different is the fact that one is a BQ view (works) and one is a BQ table (does not work).

See various screenshots.
![image](https://user-images.githubusercontent.com/93528462/183494939-48bc2147-1090-433d-b068-776259aaf07c.png)
![image](https://user-images.githubusercontent.com/93528462/183494989-8154b49f-6cfd-4b94-adb4-329e4d14bbb7.png)
![image](https://user-images.githubusercontent.com/93528462/183495072-5d471382-c410-4825-b3d5-662dded50f25.png)
![image](https://user-images.githubusercontent.com/93528462/183495438-9d8929fd-24dd-4d9f-9154-6f60a77bcc20.png)

### Expected Behavior

I expect that the table is successfully linked and all metadata is pulled back, just like when the source's underlying relation is a view.

### Steps To Reproduce

As above.

### Relevant log output

```shell
n/a
```

### Environment

```markdown
dbt: 1.2
```

### Which database adapter are you using with dbt?

bigquery

### Additional Context

n/a

jtcohen6:
@adamcunnington-mlg This is an interesting one! I've transferred to `dbt-bigquery`, since this is where the relevant code lives, and would need updating.

Here's the relevant spot in the "catalog" query that dbt uses to fill out `catalog.json` during `docs generate`, and then power the docs site:

https://github.com/dbt-labs/dbt-bigquery/blob/764f5fe6370f9f7ed941a387104f57207dd562c6/dbt/include/bigquery/macros/catalog.sql#L26-L28

[...]

PR #364, opened 2022-10-26
Using information schema

### LINKED ISSUE #113: [CT-162] Upgrade from the __tables__ construct to the information_schema.tables construct
### Describe the feature
The use of the [project_id].[dataset_id].__tables__ has been deprecated in favor of [project_id].[dataset_id].information_schema.tables. This is relevant because it is not possible to access the former with metadata only permissions (it requires getData permissions). This would allow secure doc generation and schema-only tests to be run in a lower privilege environment.

- You can grant the service account dataViewer access to all datasets managed by dbt and all datasets imported as sources. But this could be a massive over privilege.
- You can individually grant dataViewer to these specific tables but that becomes a huge headache as you end up with a grant for every new dataset and always need to keep it up to date or you'll end up with a failure.

### Additional context
None.

### Who will this benefit?
Anyone in high-security or high-compliance environments who want to utilize external dbt actors for certain tasks (github actions, for instance.)

Sure!

dbeatty10:
resolves #113

branched from https://github.com/dbt-labs/dbt-bigquery/pull/238

### Problem

As described [here](https://github.com/GeneralMills/gmi_common_dbt_utils/pull/5/files), the current implementation uses the `project.dataset.__TABLES__` metadata table which requires elevated permissions versus the information schema tables. Service accounts provisioned for dbt usage frequently do not have access to this table. This affects `dbt docs generate`.

### Proposed solution
Use `project.dataset.INFORMATION_SCHEMA.TABLES` instead. Ideally, `TABLE_STORAGE` would be used to get the number of rows + the size in logical bytes.

### Checklist


Fleid:
Hey @dbeatty10 is this `ready_for_review`? Just checking :)

Fleid:
Or do you want me to be the judge of that :D

dbeatty10:
@Fleid I think that @colin-rogers-dbt might have taken a look at this after me, but I'm not sure if he was able to get a breakthrough or ran into the same walls.

### Trade-offs
If I recall correctly, there are some trade-offs if this PR is merged as-is.

Here are the key considerations :
- Pro: `dbt docs generate` works without elevated permissions
- Con: the row count (`row_count`) will be degraded to a constant `0`
- Con: the size in bytes (`size_bytes`) will be degraded to a constant `0`

The degradations above would affect the backwards-compatibility for folks that _do_ have elevated permissions today, so `dbt docs generate` is already working fine for them right now.

### Potential paths forward

It sounds like [General Mills has a way](https://github.com/GeneralMills/gmi_common_dbt_utils/pull/5/files#diff-b8a2cba11fa5e23d49a5af4c4f49876abde8078152a1dcdba7454846ebec59f0R46-R52) to get the `row_count` and `size_bytes`, but it requires knowing the relevant `{{ region }}.INFORMATION_SCHEMA.TABLE_STORAGE` to query from.

If we can somehow infer the relevant region(s) to template within the `bigquery__get_catalog` query, then we can overcome both of the cons listed above.

Two different ideas we could try:
1. Could we try the original/current method using `project.dataset.__TABLES__` and fall-back to the new method in case of failure?
2. Could we query the list of unique `location` values within `INFORMATION_SCHEMA.SCHEMATA` and then iterate through them to generate all the unique [`region-REGION.INFORMATION_SCHEMA.TABLE_STORAGE`](https://cloud.google.com/bigquery/docs/information-schema-table-storage) queries? Then we'd have the row counts and sizes in bytes?

Fleid:
Hey @dbeatty10, it looks like you're using the better regex pattern to identify shards, like discussed [here](https://github.com/dbt-labs/dbt-bigquery/issues/260#issuecomment-1209250945).
I'm guessing that's on purpose? Would this PR solves #260?

dbeatty10:
> Hey @dbeatty10, it looks like you're using the better regex pattern to identify shards, like discussed [here](https://github.com/dbt-labs/dbt-bigquery/issues/260#issuecomment-1209250945). I'm guessing that's on purpose? Would this PR solves #260?

That's awesome @Fleid ! Credit goes to @hassan-mention-me for the regex (and basically the rest of the implementation seen in this PR). I branched from his implementation in https://github.com/dbt-labs/dbt-bigquery/pull/238, and if this PR is merged @hassan-mention-me is listed in the changelog entry and commits are preserved as well.

Giving https://github.com/dbt-labs/dbt-bigquery/issues/260 a re-read, it does look like the regex in this PR would solve it. However, I'd prefer to see explicit test cases added to confirm that the regex is working properly.

Here are some test cases listed [here](https://github.com/dbt-labs/dbt-bigquery/issues/260#issuecomment-1208582008) that should not be considered shards:
- `STD_MOBILITY_INDEXED_20220519163648`
- `foo20220808`
- `foo_bar20220808`

Fleid:
Hey @dbeatty10, do you think we can make that one not a draft, and flag it `ready_for_review`?

dbeatty10:
@Fleid there's two things that make me uncomfortable with marking this as being "ready for review", both of which I would consider a breaking change:
- the row count (row_count) will be degraded to a constant of `0`
- the size in bytes (size_bytes) will be degraded to a constant of `0`

My opinion: we should ensure that it's non-breaking first.

Here's two ideas to make it non-breaking, neither of which I have tried:
1. **Try `project.dataset.__TABLES__` and fall-back to `INFORMATION_SCHEMA.TABLES`**
    - Try accessing the original/current method using `project.dataset.__TABLES__` and fall-back to the new method in case of failure
1. **Use `INFORMATION_SCHEMA.SCHEMATA`**
    - Query the list of unique location values within `INFORMATION_SCHEMA.SCHEMATA` and then iterate through them to generate all the unique [region-REGION.INFORMATION_SCHEMA.TABLE_STORAGE](https://cloud.google.com/bigquery/docs/information-schema-table-storage) queries.

Advantage of option 1:
- Anyone that has sufficient permissions and had non-zero row counts (and sizes in bytes) before would still have non-zero values after
- People without sufficient permissions now would have `0` for row counts (and sizes in bytes), but at least they would have everything else.

Advantage of option 2:
- We'd have the row counts and sizes in bytes 100% of the time for all users.

Fleid:
I hear you.

I'm thinking that `ready_for_review` can also be about "we've tried as much as we can, but for whatever reason couldn't push it over the finish line, so let's get the team in to do it".

Do you want another stab at this, or are you good with passing the baton?

dbeatty10:
I'm marking this as `ready_for_review` to indicate that:
> we've tried as much as we can, but for whatever reason couldn't push it over the finish line, so let's get the team in to do it

I'm passing the baton to whoever reviews this! Here's the best TLDR for you to read about proposed things to resolve: https://github.com/dbt-labs/dbt-bigquery/pull/364#issuecomment-1447217693

Fleid:
Now tracking at https://github.com/dbt-labs/dbt-bigquery/issues/585

drnielsen:
If the the biggest concern is losing the row count and size, wouldn't you be able to solve that by using the INFORMATION_SCHEMA.PARTITIONS table?

Something like the below:
``` SQL
SELECT

  table_catalog as table_database,
  table_schema as table_schema,
  table_name as original_table_name,
  concat(table_catalog, '.', table_schema, '.', table_name) as relation_id,
  row_count as row_count,
  size_bytes as size_bytes,
  case when table_type = 'EXTERNAL' then 'external' else 'table' end as table_type,
  regexp_contains(table_name, '^.+[0-9]{8}$') and table_type = 'BASE TABLE' as is_date_shard,
  regexp_extract(table_name, '^(.+)[0-9]{8}$') as shard_base_name,
  regexp_extract(table_name, '^.+([0-9]{8})$') as shard_name

FROM (
  SELECT
    table_catalog,
    table_schema,
    table_name,
    table_type,
    sum(total_rows) row_count,
    sum(total_logical_bytes) size_bytes
  FROM dataset_name.INFORMATION_SCHEMA.TABLES
  LEFT OUTER JOIN dataset_name.INFORMATION_SCHEMA.PARTITIONS
  USING (table_catalog, table_schema, table_name)
  GROUP BY table_catalog, table_schema, table_name, table_type
)
```

sambloom92:
I'm new to this issue but my naive impression is that this looks like a good suggestion, are there any issues with doing it this way?

> If the the biggest concern is losing the row count and size, wouldn't you be able to solve that by using the INFORMATION_SCHEMA.PARTITIONS table?
> 
> Something like the below:
> 
> ```sql
> SELECT
> 
>   table_catalog as table_database,
>   table_schema as table_schema,
>   table_name as original_table_name,
>   concat(table_catalog, '.', table_schema, '.', table_name) as relation_id,
>   row_count as row_count,
>   size_bytes as size_bytes,
>   case when table_type = 'EXTERNAL' then 'external' else 'table' end as table_type,
>   regexp_contains(table_name, '^.+[0-9]{8}$') and table_type = 'BASE TABLE' as is_date_shard,
>   regexp_extract(table_name, '^(.+)[0-9]{8}$') as shard_base_name,
>   regexp_extract(table_name, '^.+([0-9]{8})$') as shard_name
> 
> FROM (
>   SELECT
>     table_catalog,
>     table_schema,
>     table_name,
>     table_type,
>     sum(total_rows) row_count,
>     sum(total_logical_bytes) size_bytes
>   FROM dataset_name.INFORMATION_SCHEMA.TABLES
>   LEFT OUTER JOIN dataset_name.INFORMATION_SCHEMA.PARTITIONS
>   USING (table_catalog, table_schema, table_name)
>   GROUP BY table_catalog, table_schema, table_name, table_type
> )
> ```

PR #366, opened 2022-10-26
Add aliasing method for pseudo columns

### LINKED ISSUE #365: [CT-1415] [Feature] Add adapter-specific aliasing for Google pseudo columns

### Describe the feature

**Background:** 
Google has a concept of metadata "pseudo columns" which are columns that can be queried in BigQuery, but do not materialize in the table itself. 

Some examples of these columns are `_PARTITIONDATE, _PARTITIONTIME, _FILE_NAME,  _TABLE_SUFFIX`...etc, which always start with an underscore. `_FILE_NAME` for example, will be the name of the file being read in from Google Cloud Storage when an external data source is configured.

**Issue:** 
Because these pseudo columns start with underscore, applying dbt tests to them is difficult. 

For example, dbt_utils.unique_combination_of_columns and the default dbt uniqueness test select columns as is and create the dbt__test_audit table schema based on the selected columns. 

Since columns with leading underscores are not allowed in BigQuery, applying uniqueness tests to these columns as is throws an error. The columns need to be aliased in order for test results to be stored in dbt__test_audit tables.

I've considered:

- Writing a macro within my organizations dbt project to perform this aliasing, however because the aliasing would be used in tests defined in dbt-core and external packages, this doesn't feel like the best solution. In addition, being able to test on pseudo columns will likely benefit other dbt-bigquery users.

- Aliasing pseudo columns in base models and materializing them there. Tests can then be run on the base model instead of on the source layer. The issue with that solution is we'd essentially be replicating data simply to add a uniqueness test.

### Who will this benefit?

All dbt-bigquery users that use external data sources and want uniqueness testing on their source layer.

Yes. I will be submitting a PR for a potential solution!

### Anything else?

_No response_

ashleemtib:
resolves #365 

### Description

Added the `alias_column` method to the `BigQueryAdapter` class. 

The purpose of this method is to produce a column alias for use in a select statement or completely rename a provided column. The initial use case is for this to be used to alias [pseudo columns](https://cloud.google.com/bigquery/docs/external-data-cloud-storage#the_file_name_pseudo_column) so that dbt tests can be applied to them. If not aliased, BigQuery errors as it is trying to produce the dbt__test_audit table using a column name with an illegal character (preceding _).

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/impl.py
+++ b/dbt/adapters/bigquery/impl.py
@@ -1,7 +1,6 @@
 from dataclasses import dataclass
 from typing import Dict, List, Optional, Any, Set, Union, Type
 from dbt.dataclass_schema import dbtClassMixin, ValidationError
-copy_partitions: bool = False
 
 import dbt.deprecations
 import dbt.exceptions
--- a/dbt/adapters/bigquery/impl.py
+++ b/dbt/adapters/bigquery/impl.py
@@ -544,7 +544,7 @@ class BigQueryAdapter(BaseAdapter):
                 conf_partition.field
                 if not conf_partition.time_ingestion_partitioning
                 else "_PARTITIONTIME"
-            )            
+            )
             return (
                 table_field == conf_table_field.lower()
                 and table_granularity == conf_partition.granularity.lower()
--- a/dbt/adapters/bigquery/impl.py
+++ b/dbt/adapters/bigquery/impl.py
@@ -67,6 +67,7 @@ class PartitionConfig(dbtClassMixin):
     granularity: str = "day"
     range: Optional[Dict[str, Any]] = None
     time_ingestion_partitioning: bool = False
+    copy_partitions: bool = False
 
     def reject_partition_field_column(self, columns: List[Any]) -> List[str]:
         return [c for c in columns if not c.name.upper() == self.field.upper()]
--- a/dbt/adapters/bigquery/impl.py
+++ b/dbt/adapters/bigquery/impl.py
@@ -248,8 +248,8 @@ class BigQueryAdapter(BaseAdapter):
             return []
 
     @available
-    def reformat_column_name(self, column) -> str:
-        "Rename pseudo columns to remove leading underscore"
+    def alias_column_name(self, column) -> str:
+        "Alias pseudo columns to remove leading underscore"
         if column[0] == '_':
             column_alias = column.replace('_','',1)
             column = column + f' as {column_alias}'
--- a/dbt/adapters/bigquery/impl.py
+++ b/dbt/adapters/bigquery/impl.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
 from typing import Dict, List, Optional, Any, Set, Union, Type
 from dbt.dataclass_schema import dbtClassMixin, ValidationError
+copy_partitions: bool = False
 
 import dbt.deprecations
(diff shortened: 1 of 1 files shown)


ashleemtib:
Still working on checklist!

joellabes:
I won't actually be reviewing the PR, but am going to approve the workflows running to get CI things kicked off for you in the meantime!

dbt/adapters/bigquery/impl.py
```diff
@@ -248,6 +248,17 @@ def get_columns_in_relation(self, relation: BigQueryRelation) -> List[BigQueryCo
             logger.debug("get_columns_in_relation error: {}".format(e))
             return []
 
+    @available
+    def alias_column(self, column) -> Dict[str, str]:
```
jtcohen6, line 252: Where would we be checking to see if a column name needs aliasing, on account of starting with `__`? Are you thinking that we'd add calls to `alias_column` within the code that passes the `column_name` value into generic test macro definitions?

jtcohen6 (commented):
Thanks for identifying the problem, @ashleemtib, and opening the PR to fix it!

I'm curious — what do you envision the end user experience looking like? Should I be able to define a test on a pseudo-column like:
```
models:
  config:
    partition_by: ...
  columns:
    - name: _PARTITIONDATE
      tests:
        - unique
```
And the `unique` test would just work, because the aliasing (`_PARTITIONDATE as partition_date`) would be handled implicitly?

dbt/adapters/bigquery/impl.py
ashleemtib: Hi Jeremy, thanks for the response and questions! Your example of how the test would be defined in the yml is correct for what I was thinking for the user experience. 

As for checking if the column name needs aliasing, yes, the idea was that generic tests could call this method and remove the leading underscore. It may need to be done differently depending on the macro which is why I have it returning a dictionary of the original column name, the aliased column name, or the actual aliasing that would need to be included in a select statement: `select _FILE_NAME as FILE_NAME`. Here is an example of how I updated `dbt_utils.unique_combination_of_columns` to use the method: 

```
{% macro bigquery__test_unique_combination_of_columns(model, combination_of_columns, quote_columns=false) %}

{%- set select_list=[] %}
{% if not quote_columns %}
        {% set group_list = combination_of_columns %}
        {% for column in combination_of_columns -%}
            {% set column = adapter.alias_column( column )['statement'] or column %}
            {% set select_list = select_list.append( column ) %}
        {%- endfor %}
{% elif quote_columns %}
        {% for column in combination_of_columns -%}
            {% set select_list = select_list.append( adapter.quote(column) ) %}
            {% set group_list = select_list %}
        {%- endfor %}
{% else %}
    {{ exceptions.raise_compiler_error(
        "`quote_columns` argument for unique_combination_of_columns test must be one of [True, False] Got: '" ~ quote ~"'.'"
    ) }}
{% endif %}

{%- set group_csv, select_csv=group_list | join(', '), select_list | join(', ') %}


with validation_errors as (

    select
        {{ select_csv }}
    from {{ model }}
    group by {{ group_csv }}
    having count(*) > 1

)

select *
from validation_errors

{% endmacro %}
```

Fleid:
I commented on the original issue last week, waiting for an answer to decide how to move forward on that PR.

Fleid:
Closed as discussed in the issue.

Issue #16, opened 2020-03-30
New insert_overwrite Bigquery partitioning with integer keys can create huge temporary array variables, exceeding BQ limits


haukeduden:
### Describe the bug
When doing an incremental update of an integer-partitioned bigquery table with the new insert_overwrite merge strategy then DBT calculates which partitions it should replace. In the process it can generate a huge BigQuery ARRAY value.

The cause is that DBT does not take the "interval" parameter of the partitioning specification into account. The generated SQL code selects "array_agg(distinct PARTITION_FIELD)" when calculating the partitions to replace. This selects ALL distinct partition field values of the incremental update, even if these values are actually in the same partition. This causes a potentially huge array to be created. If there is enough data in the table then this will even cause an error because BQ's maximum array size (100 MB) is exceeded.

Note that this bug is not triggered by time partitioned tables because for those all partition fields are dates and the partition size is always one day (i.e. there is only one valid value per partition).

### Steps To Reproduce
1) Generate a model with:
config(
    materialized = 'incremental',
    unique_key = 'partkey',

    partition_by = {
        "field": "partkey",
        "data_type": "int64",
    
        "range": {
            "start": 0,
            "end": 100000,
            "interval": 1000
        }
    },
   incremental_strategy = 'insert_overwrite'
  )

2) Run the model once with empty source data

3) Add 1000 records in the source data with values for partkey from 0 to 999

4) Generate the incremental update code for the model and look at the values in the dbt_partitions_for_replacement internal variable.

5) You will see that it contains all 1000 values for partkey from 0 to 999, even though those are all inside the same single partition.

### Expected behavior
DBT should ensure that no huge temporary array variable is generated.
The dbt_partitions_for_replacement array should have at most as many elements as the number of partitions being updated. In my opinion the way to go would be to store only the starting values of each partition in the array and then modify the merge clause to use a range for each partition.

### Screenshots and log output
N/A

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [x] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
installed version: 0.16.0
   latest version: 0.16.0

Up to date!
```

**The operating system you're using:**

macOS

**The output of `python --version`:**
Python 3.7.4

judahrand:
[...]

PR #371, opened 2022-11-02
Subvert `insert_overwrite` merge strategy to bring back merge

elyobo:
Resolves https://github.com/dbt-labs/dbt-bigquery/issues/16

### Description

For comment only to see if there's any interest in the idea, not necessarily this particular implementation.

The insert overwrite implementation uses BigQuery scripting variables to extract store the partition values that will be updated, allowing efficient pruning of the insert query with substantial cost and performance benefits. This works well when those values are small enough to fit inside BQ scripting's memory limit of 1 MiB as described in https://github.com/dbt-labs/dbt-bigquery/issues/16, which I guess is going to be hit at 16,384 INT64 IDs.

Prior to the current insert overwrite strategy, @jtcohen6 worked on a similar approach that used minimum and maximum partition values instead in https://github.com/dbt-labs/dbt-core/pull/1971, which works around the memory limit because only two values will be selected. It's not as efficient (unnecessary partitions between the min and the max will still be read), but the worst case performance is that it's no better than the merge + temp table cost (in which case users can opt to use the merge strategy) and in the best case (a contiguous block of IDs, e.g. in append only data) then it will prune partitions as well as the insert overwrite.

This PR just tinkers with the insert overwrite to bring back this approach as an optional variant, but it's probably confusing UX to have this very-much-not-an-insert-overwrite inside the "insert overwrite" strategy, and it has no tests - I made it to confirm that it was possible and that it worked in practice not just in my head, and it did offer significant improvements for me.

If there's interest in bringing this back I'm happy to work on an improved version that addresses these and any other issues raised.

### Alternatives Considered

Users can also implement this themselves (to a large extent, at least; I had SQL header issues) by implementing their own `bigquery__get_merge_sql`. I added one in https://github.com/dbt-labs/dbt-bigquery/issues/16#issuecomment-1296028981 as an example which is working for me although, as mentioned, I had SQL header issues - they're output to build the temp table and I can't seem to suppress them, so they're output again for the real select.

It would be more convenient, simple, and less error prone if this was supported directly.

Users can also fall back to the existing merge approach, but this can be much more expensive.

### Checklist

### DIFF
--- a/dbt/include/bigquery/macros/materializations/incremental_strategy/insert_overwrite.sql
+++ b/dbt/include/bigquery/macros/materializations/incremental_strategy/insert_overwrite.sql
@@ -104,9 +104,7 @@
   {%- if copy_partitions is true %}
      {{ bq_dynamic_copy_partitions_insert_overwrite_sql(tmp_relation, target_relation, sql, unique_key, partition_by, dest_columns, tmp_relation_exists, copy_partitions) }}
   {% else -%}
-      {% set predicate -%}
-          {{ partition_by.render_wrapped(alias='DBT_INTERNAL_DEST') }} in unnest(dbt_partitions_for_replacement)
-      {%- endset %}
+      {%- set is_exact_match_strategy = config.get('match_strategy', 'exact') == 'exact' -%}
 
       {%- set source_sql -%}
       (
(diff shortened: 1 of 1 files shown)


elyobo:
Happy to sign the CLA if there's interest in getting this merged bot :) I did have [a minor PR in `dbt-utils`](https://github.com/dbt-labs/dbt-utils/pull/686), different CLA requirements there?

elyobo:
Would appreciate any feedback on the idea here; it's a valuable cost saver for BigQuery users where the insert/overwrite strategy doesn't work due to memory limits in BQ "scripts"  e.g. as described in https://github.com/dbt-labs/dbt-bigquery/issues/16.

elyobo:
Implementation being discussed in https://github.com/dbt-labs/dbt-bigquery/issues/16 but this doesn't look like the way we'll be going so closing.

PR #379, opened 2022-11-07
[CT-1412] proposed fix to override_database functional test hanging artifacts

### LINKED ISSUE #360: [CT-1412] Fix hanging tests in dbt-bigquery during testing

### Current Behavior

during testing some tests that have a change to either `schema` or `database` name change are not having those datasets/tables fully torn down at end of test, slowly filling out the testing projects in bigquery leading to nuisance failures.

### Expected Behavior

all tests should have temp generated datasets or tables fully torn down leaving a clean testing smaller footprint in the testing project

A change to the `dbt-bigquery` version of `drop_schema` should be able to fix this issue.

### Steps To Reproduce

Run these tests locally below check database, or check for nightly ci/cd runs

Tests that cause this error often are:

1. The `defer_state` integration test we leave behind `defer_state_other` because the recognized relation used during `drop_schema` is `dbt-test-env`.`test16656003516500320071_defer_state` where as the relation name in results from `results = self.run_dbt(['run', '-m', 'view_model', '--state', 'state', '--defer', '--target', 'otherschema'])`  is `'`dbt-test-env`.`test16656003516500320071_defer_state_other`.`view_model`'` .
2. During the `override_database` functional test for classes like `TestProjectModelOverrideBigQuery` we change the database part of schema for several models this would differ from what `drop_schema` would go to drop.
3. `Docs_generate` changes schema on second_model from `test16666427620405249475_test_basic` to `test16666427620405249475_test_basic_test` by changing schema name in the model config which means when we trigger `drop_schema` it isn't looking for the correct schema

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-bigquery:
```

### Additional Context

_No response_

McKnight-42:
resolves #360 

### Description

add a new method in class `BaseOverrideDatabase` to add a extra `drop_schema` targeting a `relation` pointing to alternate project.

spike doc: https://www.notion.so/dbtlabs/Bigquery-Leftover-tests-artifacts-350aa89a047f44b3b3b1967aaff8367e

todo:
- [ ] backport to 1.3.latest
- [ ] backport to 1.2.latest
- [ ] backport to 1.1.latest

### Checklist


dbeatty10:
@McKnight-42 With https://github.com/dbt-labs/dbt-bigquery/issues/360 resolved, is this pull request still relevant? Or can it be closed?

McKnight-42:
@dbeatty10 This is one of 3 notable tests that was causing #360 I just merged in the initial pr for `defer_state` which likely closed out the issue, the backports for that one are up, waiting on the fix from today to be merged into 1.1.latest so I can merge in the docs_generate https://github.com/dbt-labs/dbt-bigquery/pull/373 initial pr and still waiting on review for this one

dbeatty10:
@McKnight-42 sounds good 👍 

Do we have a pattern or workflow to avoid prematurely closing issues? For example, by using a convention like "related to #360" rather than "resolves #360" if there are multiple PRs involved?

tests/functional/test_override_database/test_override_database.py
```diff
@@ -31,6 +31,10 @@ def project_config_update(self):
             }
         }
 
+    def delete_alt_database_relation(self, project):
```
mikealfare, line 34: If this is what we always want to do, and ensure that it runs regardless of the success of the unit test, can we run it as a teardown method?

McKnight-42: My initial design was for it to be a `fixture` or called a `tearDown` method but felt it might distract people in future troubleshooting cases as this method is only used to cover the small case that our functional tests default method of `drop_test_schema` misses which is renames or pointing to different projects from our default relation information. this really only happens in a very few tests this one, and `tests_aliases` being a notable.

the `fixture` didn't seem to trigger properly and would leave our hanging artifacts in the alt bigquery project, this version seems to work. so just figured to explicitly name the method based on what its needed for

mikealfare: Ah I see. And this is where I'm lacking the context. My spidey sense is saying that we should look into the few cases where it doesn't work as expected then. But I agree, if it works most of the time, we should not run this as a teardown for all cases.

VersusFacit: I'd still prefer to see this as function 😁 It's a pretty quick thing to add a fixture to the class, and then, you could call it within the function definition of your test case wherever this behavior should be forced. See the `yield` statement in the fixtures docs -- Chenyu turned me onto this. I suspect this will appear in other tests and having it so easily accessible would be rad to the max

VersusFacit (changes requested):
Let's try out the method mike suggested for this test as opposed to manually invoking the function. In other words, make a fixture on for this class (inside the class) that you call in the function definitions of other test functions. Alternatively, override the tearDown fixture we have.

If you want an example, seek a test in core with `yield`.

tests/functional/test_override_database/test_override_database.py
McKnight-42: played around and was able to get a fixture built out

tests/functional/test_override_database/test_override_database.py
```diff
@@ -31,6 +29,12 @@ def project_config_update(self):
             }
         }
 
+    @pytest.fixture(scope="class")
+    def clear_test_schema(self, project):
+        yield
+        relation = project.adapter.Relation.create(database=os.getenv("BIGQUERY_TEST_ALT_DATABASE"), schema=project.test_schema)
```
colin-rogers-dbt, line 35: Looks like we call `os.getenv("BIGQUERY_TEST_ALT_DATABASE")` in a bunch of places, could we just assign this to a static var?

McKnight-42: sure

colin-rogers-dbt (approved):
LGTM

PR #359, opened 2022-10-25
Make dataset updates single threaded

colin-rogers-dbt:
### Description
Experimental attempt at solving #266 

Specifically:
* don't allow adapter to call dataset update in parallel
* don't block threads on dataset updates

### Checklist

- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #404, opened 2022-12-02
add dataset update lock

### LINKED ISSUE #266: [CT-1032] `grant_access_to` only works reliably in single threaded mode
### Describe the bug
A clear and concise description of what the bug is. What command did you run? What happened?

### Steps To Reproduce
In as much detail as possible, please provide steps to reproduce the issue. Sample data that triggers the issue, example model code, etc is all very helpful here.

If applicable, add screenshots or log output to help explain your problem.

### System information
```
<output goes here>
```

colin-rogers-dbt:
resolves #266 
### Description
A simpler implementation than #359 this preserves the expected behavior of how model execution happens (i.e. don't continue executing child models if the grant access call fails.)

### Checklist


dbt/adapters/bigquery/impl.py
```diff
@@ -805,27 +807,28 @@ def grant_access_to(self, entity, entity_type, role, grant_target_dict):
...
+            if access_entry in access_entries:
+                logger.debug(f"Access entry {access_entry} " f"already exists in dataset")
+                return
 
-        access_entries.append(AccessEntry(role, entity_type, entity))
-        dataset.access_entries = access_entries
-        client.update_dataset(dataset, ["access_entries"])
+            access_entries.append(AccessEntry(role, entity_type, entity))
+            dataset.access_entries = access_entries
+            client.update_dataset(dataset, ["access_entries"])
```
dataders, line 831: can you ELI5 this solution?

colin-rogers-dbt: it declares a global lock (i.e. all threads share it) and then whenever a thread hits this function it attempts to "acquire" the lock and waits until it can before continuing. This means only one thread can execute this code block at a time.

dbt/adapters/bigquery/impl.py
```diff
@@ -805,27 +807,28 @@ def grant_access_to(self, entity, entity_type, role, grant_target_dict):
...
-            logger.debug(f"Access entry {access_entry} " f"already exists in dataset")
-            return
+            if access_entry in access_entries:
+                logger.debug(f"Access entry {access_entry} " f"already exists in dataset")
+                return
 
-        access_entries.append(AccessEntry(role, entity_type, entity))
-        dataset.access_entries = access_entries
-        client.update_dataset(dataset, ["access_entries"])
+            access_entries.append(AccessEntry(role, entity_type, entity))
```
mikealfare, line 829: access_entries.append(access_entry)

mikealfare: Unless the access_entry assignment needs to be in the lock, I would move that assignment outside of the lock, and then make the contents of the `with` block a separate function, e.g. `self._update_access_entries(access_entry, grant_target)`.

colin-rogers-dbt: since access entries is tied to the dataset being updated I believe we do actually need to perform this logic as part of a transaction within the lock: aquire lock>get the dataset>get the access entries for the dataset>check if the access entry exists>update the dataset>exit lock

dbt/adapters/bigquery/impl.py
```diff
@@ -807,26 +807,25 @@ def grant_access_to(self, entity, entity_type, role, grant_target_dict):
...
             access_entries = dataset.access_entries
 
+            logger.warning(f"Access entries: {access_entries}")
             if access_entry in access_entries:
-                logger.debug(f"Access entry {access_entry} " f"already exists in dataset")
+                logger.warning(f"Access entry {access_entry} " f"already exists in dataset")
                 return
 
-            access_entries.append(AccessEntry(role, entity_type, entity))
+            access_entries.append(access_entry)
```
mikealfare, line 828: This `access_entry` is not the one from line 820, but instead is the last one in `access_entries` from line 824. Is that intended?

tests/unit/test_bigquery_adapter.py
```diff
@@ -996,18 +996,17 @@ def test_grant_access_to_does_not_call_update_with_duplicate_access_entry(self):
...
         a_different_entity = BigQueryRelation.from_dict(
-        {
-            "type": None,
-            "path": {"database": "another-test-project", "schema": "test_schema_2", "identifier": "my_view"},
-            "quote_policy": {"identifier": True},
-        }
-    )
+            {
+                "type": None,
+                "path": {"database": "another-test-project",
```
mikealfare, line 1005: This is minor. I would either put path on one line (but not if it's longer than 80 char):
```"path": {"database": "another-test-project", "schema": "test_schema_2", "identifier": "my_view"},```
or put an entry per line:
```
"path": {
    "database": "another-test-project",
    "schema": "test_schema_2",
    "identifier": "my_view"
}
```

File: tests/functional/adapter/test_grant_access_to.py [hunk 5]
```diff
@@ -0,0 +1,40 @@
...
+    def models(self):
+        return {
+            "select_1.sql": SELECT_1,
+            "select_1_table.sql": SELECT_1_TABLE,
+        }
+
+
+class TestAccessGrantSucceeds(GrantAccessToConfig):
+    def test_grant_access_succeeds(self, project):
+        #Need to run twice to validate idempotency
```
mikealfare, line 37: Should we `assert len(results) == 2` after each `run_dbt()` call then? In other words, add one between line 38 and line 39.

dbt/adapters/bigquery/dataset.py
```diff
@@ -0,0 +1,35 @@
...
+    # we can't simply check if an access entry is in the list as the current equality check
+    # does not work because the locally created AccessEntry can have extra properties.
+    for existing_entry in access_entries:
+        role_match = existing_entry.role == access_entry.role
+        entity_type_match = existing_entry.entity_type == access_entry.entity_type
+        if role_match and entity_type_match:
+            new_prop = access_entry._properties
+            existing_prop = existing_entry._properties
+            properties_match = [new_prop[k] == v for k,v in existing_prop.items()]
+            if properties_match:
```
mikealfare, line 30: Perhaps I'm missing something, but can we simplify lines 27-30:
```
if access_entry._properties == existing_entry._properties:
```

If so, we combine this with lines 24 and 25 as well:
```
for existing_entry in access_entries:
    if all(
        existing_entry.role == access_entry.role,
        existing_entry.entity_type == access_entry.entity_type,
        existing_entry._properties == access_entry._properties
    ):
        logger.warning(f"Access entry {access_entry} already exists in dataset")
        return dataset
```

And then if we want to be really fancy, we can define ```def __eq__(self)``` on AccessEntry, which turns this whole function into:
```
if access_entry in dataset.access_entries:
    logger.warning(f"Access entry {access_entry} already exists in dataset")
    return dataset
access_entries.append(access_entry)
dataset.access_entries = access_entries
return dataset
```
which can be simplified further to:
```
dataset.access_entries = list(set(dataset.access_entries).add(access_entry))
return dataset
```

And if we can change the datatype of dataset.access_entries, then this whole function can be avoided as it's equivalent to:
```
dataset.access_entries.add(access_entry)
```

colin-rogers-dbt: this would definitely be the right way to solve this, the issue is that the `_properties` fields do not match as there appears to be a difference in how BQ stores/returns access entries versus how they are locally constructed. The result is the local version can have extra attributes in its `_properties`.

The right long term thing to do here is to flag this issue to BQ.

tests/functional/adapter/test_grant_access_to.py
colin-rogers-dbt: good call! added

tests/functional/adapter/test_grant_access_to.py
```diff
@@ -0,0 +1,41 @@
...
+class GrantAccessToConfig:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "select_1.sql": SELECT_1,
+            "select_1_table.sql": SELECT_1_TABLE,
+        }
+
+
+class TestAccessGrantSucceeds(GrantAccessToConfig):
```
nathaniel-may, line 35: could we add a test for this feature failing properly? i.e. assert that child models aren't run after a bad grant statement.

tests/unit/test_dataset.py
```diff
@@ -0,0 +1,23 @@
...
+            "path": {"database": "test-project", "schema": "test_schema", "identifier": "my_view"},
+            "quote_policy": {"identifier": False},
+        }
+    )
+    dataset_ref = DatasetReference(project=database, dataset_id=dataset)
+    dataset = Dataset(dataset_ref)
+    access_entry = AccessEntry(None, "table", entity)
+    dataset = add_access_entry_to_dataset(dataset, access_entry)
+    assert access_entry in dataset.access_entries
+    dataset = add_access_entry_to_dataset(dataset, access_entry)
```
mikealfare, line 22: Because of the _properties issue called out in the other file, should we create a separate instance of AccessEntry with the same input instead? We want `add_access_entry_to_dataset()` to only add one access entry based on its properties, not the instance equality.

dbt/adapters/bigquery/dataset.py
```diff
@@ -20,15 +20,9 @@ def add_access_entry_to_dataset(dataset: Dataset, access_entry: AccessEntry) ->
     # we can't simply check if an access entry is in the list as the current equality check
     # does not work because the locally created AccessEntry can have extra properties.
     for existing_entry in access_entries:
-        role_match = existing_entry.role == access_entry.role
```
mikealfare: It looks like we dropped the check on `.role` and `.entity_type`. Are those covered by checking `._properties`?

tests/functional/adapter/test_grant_access_to.py
```diff
@@ -21,21 +22,47 @@
...
+        }
+
+    def test_grant_access_fails_without_running_child_table(self, project):
+        # Need to run twice to validate idempotency
+        results = run_dbt(["run"], expect_pass=False)
+        assert results[0].status == "error"
+        assert results[1].status == "skipped"
+        assert (
+            results[0].message
+            == "404 GET https://bigquery.googleapis.com/bigquery/v2/projects/dbt-test-env-alt/datasets/NonExistentDataset?prettyPrint=false: Not found: Dataset dbt-test-env-alt:NonExistentDataset"
```
mikealfare, line 67: Do we specifically want to enforce this entire string? It looks very specific, which suggests it might also be brittle. We might wind up with a failed test in the future that doesn't reflect broken functionality. Can we loosen the constraint to something like:
```
assert results[0].message.startswith("404 GET https://bigquery.googleapis.com/")
```

tests/functional/adapter/test_grant_access_to.py
colin-rogers-dbt: Added!

dbt/adapters/bigquery/dataset.py
colin-rogers-dbt: turns out they are not covered by properties

mikealfare (approved):

PR #449, opened 2022-12-23
Create new materialization, Gated Table Materialization, that allows models to split based on config logic. 

### LINKED ISSUE #450: [CT-1726] [Feature] Creating a new table materialization that separates data into good/bad

### Describe the feature

Instead of using the `table` materialization, use a materialization called `gated_table` (name open to interpretation). 

This materialization accepts an additional item in the config block, let's call it `gating_logic`, which is a list of all the WHERE clause statements you'd like to apply to your model. These statements are then interpolated into your model, and the materialization outputs two mutually exclusive models: one takes in every row that passes all the gating logic statements, the other table takes in any row that failed any of the gating logic statements.

Why? This allows you to keep track of data failures. Instead of applying the filters within the model as WHERE CLAUSE statements, which erases the data from downstream models, it just splits it into a separate table. This failure table has a common table prefix, so that you can use BigQuery wildcard syntax to split out and aggregate all the failure tables, dashboarding them / record keeping with them. 

The alternative is setting up your tables in a structure where you create a separate pass through view that only has the good data, but for users/groups that have already set up their tables with certain conventions, it's especially hard to go ahead and make these changes which would create hundreds if not thousands of additional tables/views. 

In the above scenario, you could then apply singular tests or YML tests to the base table, to keep track of all the failures. This is probably a bit more dbt-esque, but the issue is that refactoring an existing database to follow this convention is pretty hard and relies on the creation of a ton of tables. Furthermore, it makes updates a bit harder since you need to add / change columns on multiple tables, versus just one table that automatically splits. 

### Who will this benefit?

This will benefit anyone who is looking to have really solid "final" tables that only have good data, but also want the ability to fix their data / keep track of data issues in the future. You could also modify the naming conventions and have this just be a general table split - by using a WHERE clause you can separate tables out into two separate versions, and then maybe supply the table naming convention as an extra config parameter as well. 

Yes! I have already written the code for the materialization and tested it out on a PoC. 

### Anything else?

I've already wrote the code and an example PoC use case. Happy to share that with y'all. I just need some help figuring out the testing scenarios / bulletproofing the materialization logic so that it works!

noahfkennedy:
resolves #450 

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [X] I have run this code in development and it appears to resolve the stated issue
- [X] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR 
```
I don't think docs are required for this change, based on the addition of another macro incremental materialization, but happy to write docs if someone disagrees. 
```
- [X] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)


noahfkennedy:
In the process of writing tests. I'm not sure if this requires document updates. Really would love any support / suggestions to get this over the finish line, I think that it could be very helpful to a lot of different parties.

Ultimately it satisfies one major issue: Let's say you want to filter out bad data, but you need a record of the bad data too (theoretically so you can fix it). 

If you have data quality problems, filtering rows out using CASE/WHEN statements erases the issue entirely, but you never have the chance to see bad data to begin with! Then you can't fix it. Alternatively, you can create multiple models/views, but if you're already working with a mature data hierarchy (like my team at Tempus), then it's pretty hard to make another 500+ models (time consuming, really messes up your existing repositories, requires lots of restructuring). 

This materialization allows you to just change the config block, and then it outputs two separate tables. One with good data, one with bad data.

dbeatty10:
Closing per discussion in https://github.com/dbt-labs/dbt-bigquery/issues/450

PR #476, opened 2023-01-24
[CT-1630] Convert column_types test for dbt-bigquery

McKnight-42:
resolves #https://github.com/dbt-labs/dbt-core/issues/6404

### Description

convesion of 056 to new functional test method

todo before merge: revert dev_requirements dbt-core pin

### Checklist


colin-rogers-dbt (approved):

tests/functional/adapter/column_types/test_column_types.py
```diff
@@ -0,0 +1,34 @@
+import pytest
```
mikealfare, line 1: Is there a reason to separate this from `test_alter_column_types.py`? The two fixtures (module level variables) are copies of fixtures in the other file. So the only new content is the class with two methods.

I would keep the name `test_column_types.py` as it's more generic, so I guess I'm saying to move the other file's contents here technically.

McKnight-42: These are separated for now as it is only currently tested in bigquery and is how it was historically, this does also raise a fine question of how big of a change in functionality a test is doing should they be before they are their own file/test but still grouped together.

VersusFacit: I'd suggest just revising this to have the fixtures in the same file, then import them, thereby reducing the duplicate code. Best of both worlds. 

Nit: do that

VersusFacit (approved):

tests/functional/adapter/column_types/test_column_types.py
McKnight-42: There are a few very slight differences in the fixtures for each test, one easily fixed by a project_config_update in that we materialize alt model as a table, the second for the schema being a change in the int64 type

PR #482, opened 2023-01-26
Use IEC standard abbreviations (GiB, TiB, etc)

### LINKED ISSUE #477: [CT-1892] [Bug] Non-standard abbreviations for kibibyte, etc

### Context
[IEC 80000-13](https://en.wikipedia.org/wiki/ISO/IEC_80000#Units_of_the_ISO_and_IEC_80000_series)

> Binary units of measurement express the size of data more accurately. When you compare the size of 100 KB to 100 KiB, the difference is relatively small, 2.35%. However, this difference grows as the size of the data values increases. When you compare the size of 100 TB to 100 TiB, the difference is 9.06% [[1](https://www.ibm.com/docs/en/spectrum-control/5.4.2?topic=concepts-units-measurement-storage-data)].

image

### Current Behavior

Current code:
https://github.com/dbt-labs/dbt-bigquery/blob/3ce88d75ac78b9aac1dcaa184c94e235f661a6f0/dbt/adapters/bigquery/connections.py#L258-L269

### Expected Behavior

If we use logic like `num_bytes /= 1024.0`, then I'd expect it to align with [IEC 80000-13](https://en.wikipedia.org/wiki/ISO/IEC_80000#Units_of_the_ISO_and_IEC_80000_series) and use the following standardized abbreviations instead:
```python
for unit in ["Bytes", "KiB", "MiB", "GiB", "TiB", "PiB"]:
```

(See the chart [here](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units) for where I got the official abbreviations from.)

Alternatively, the current abbreviations for decimal multi-byte units could be kept and these substitutions could be made instead:
```python
num_bytes /= 1000.0
```
and:
```python
num_bytes *= 1000.0
```

### Steps To Reproduce

Didn't actually reproduce -- just read the code!

### Relevant log output

_No response_

### Environment

```markdown
- OS: N/A
- Python: N/A
- dbt-core: >= 1.0.0
- dbt-bigquery: >= 1.0.0
```

### Additional Context

`label: tech_debt`

Today I Learned (TIL), `format_bytes` and `format_rows_number` are defined in dbt-core!
https://github.com/dbt-labs/dbt-core/blob/7b464b8a4957ec7969f19234020e110be1987923/core/dbt/utils.py#L456-L473

#### Context for the duplication
Removing them from dbt-core was mentioned [here](https://github.com/dbt-labs/dbt-bigquery/pull/48#discussion_r742152754) and [here](https://github.com/dbt-labs/dbt-bigquery/pull/48#discussion_r743485981) as part of the discussion for #48.

It feels to me like one of the following should have happened along with https://github.com/dbt-labs/dbt-bigquery/pull/48:
1. a simultaneous PR should have removed `format_bytes` and `format_rows_number` from dbt-core; or
2. [#48](https://github.com/dbt-labs/dbt-bigquery/pull/48) should have been added to dbt-core instead of dbt-bigquery.

As it stands, the implementations in dbt-core are unused (to the best that we can tell), but would probably need to be maintained to preserve backwards compatibility in case some unknown adapter is using them.

#### Proposal

Going forward, I'd propose that we choose one (and only one) of the following:
1. Keep separate implementations in dbt-core and dbt-bigquery (and fix the KB vs. KiB issue in both)
1. Remove the implementations in dbt-core (so they exist only in dbt-bigquery)
1. Completely move the implementation in dbt-bigquery back into dbt-core

dbeatty10:
resolves #477
see also https://github.com/dbt-labs/dbt-core/pull/6742

### Description

When using multiples of 1024, the IEC standard abbreviations are "GiB", "TiB", "PiB" instead of "GB", "TB", "PB" (which are for multiples of 1000).

### Checklist


Fleid:
Hey @dbeatty10 do we expect any side effects here? Or is it "just" about the log format?

dbeatty10:
> Hey @dbeatty10 do we expect any side effects here? Or is it "just" about the log format?

I'm not aware of any side effects -- I think its "just" about the log format! My understanding is that these units are for human consumption rather than machine-readable values.

Here's an example screenshot from the original implementation in https://github.com/dbt-labs/dbt-core/pull/2530#issuecomment-642902939

image

Fleid (approved):

PR #485, opened 2023-01-27
Fix time ingestion partitioning option regression

### LINKED ISSUE #483: [CT-1912] [Regression] `require_partition_filter` usage with `insert_overwrite` fails on second run

### Current Behavior

Let's take this simple example:
```
{{
    config(
        materialized = 'incremental',
        incremental_strategy='insert_overwrite',
        partition_by = {
          "field": "hour",
          "data_type": "timestamp",
          "granularity": "hour"
        },
        require_partition_filter = true,
    )
}}
SELECT TIMESTAMP('2023-01-25') as hour, 1 as value
```

If you run it twice it will fail with `"Query error: Cannot query over table '<project>.<dataset>.test_bug' without a filter over column(s) 'hour' that can be used for partition elimination at [39:5]"`

The query should be successful.

### Steps To Reproduce

Using `dbt-bigquery 1.4.0rc1 (or any 1.4.0)`

Why it fails?
Because the generated code is the following 

```
...
when not matched by source
         and timestamp(timestamp_trunc(DBT_INTERNAL_DEST.hour, hour)) in unnest(dbt_partitions_for_replacement)
        then delete
...
```

instead of

```
...
when not matched by source
         and timestamp_trunc(DBT_INTERNAL_DEST.hour, hour) in unnest(dbt_partitions_for_replacement)
        then delete
...
```

The reason is that BigQuery can't figure out `timestamp(timestamp_trunc(DBT_INTERNAL_DEST.hour, hour))` is the same as `timestamp_trunc(DBT_INTERNAL_DEST.hour, hour)` which means it can't forward the origin of the field (`hour`) to the resulting recreated field. Though we could expect the query planner to do job of inline (and prune) the timestamp cast... it's not the case so we'll have to be smarter when we generate the query to not include it.

### Workaround

Use `"copy_partitions": true` setting such as
```
 partition_by = {
          "field": "hour",
          "data_type": "timestamp",
          "granularity": "hour",
          "copy_partitions": true
        }
```
in 

### Relevant log output

_No response_

### Environment

```markdown
- OS:Mac 12.6
- Python: 3.10.9
- dbt-core (working version):1.3.2
- dbt-bigquery (working version): 1.3.0
- dbt-core (regression version): 1.4.0-X
- dbt-bigquery (regression version): 1.4.0-X
```

### Additional Context

_No response_

Kayrnt:
resolves #483 

### Description

To fix the wrap issue that prevent BQ to detect the type for partition pruning, I'm checking if we're going to wrap in the render to avoid doing it "twice".

### Checklist


anaghshineh:
Thank you, @Kayrnt! A question - any chance we could get at least one test for the situation where we're using `insert_overwrite`, with `require_partition_filter=true`, but `time_ingestion_partitioning=false`?

Another, broader question - should we consider adding `require_partition_filter=true` to the other `insert_overwrite`-related test cases? Should still allow us to test for the same functionality, but just checking the validity of the partition pruning that is already (ostensibly) happening. Thanks!

github-christophe-oudar:
>Thank you, @Kayrnt! A question - any chance we could get at least one test for the situation where we're using insert_overwrite, with require_partition_filter=true, but time_ingestion_partitioning=false?
I think it's already the case with [that test](
https://github.com/dbt-labs/dbt-bigquery/blob/7f02de1b43c5b4a99f6cd21ead3efa74508c4f3c/tests/integration/bigquery_test/test_bigquery_changing_partitions.py#L108)

Testing is the actual behavior might be a bit trickier but might be done by checking billed bytes I guess and also failing if one model doesn't filter the partition but is allowed to run.

anaghshineh:
> > Thank you, @Kayrnt! A question - any chance we could get at least one test for the situation where we're using insert_overwrite, with require_partition_filter=true, but time_ingestion_partitioning=false?
> > I think it's already the case with [that test](https://github.com/dbt-labs/dbt-bigquery/blob/7f02de1b43c5b4a99f6cd21ead3efa74508c4f3c/tests/integration/bigquery_test/test_bigquery_changing_partitions.py#L108)
> 
> Testing is the actual behavior might be a bit trickier but might be done by checking billed bytes I guess and also failing if one model doesn't filter the partition but is allowed to run.

Hmm I don't think that test you linked to is testing for what I'm talking about. Also, seems like there was refactoring done around tests in the repo - the associated file lives [here](https://github.com/dbt-labs/dbt-bigquery/blob/main/tests/functional/test_changing_partitions.py) now. Those tests are working off of a model with `table` materialization - not an incremental model.

github-christophe-oudar:
Indeed my local fork was outdated.
You're right that it doesn't look like `incremental` & `require_partition_filter` with column type partitioning is tested.

colin-rogers-dbt:
Running the tests locally it looks like `test.incremental_overwrite_day_with_time_ingestion` fixture results in SQL that's missing declaring the `_dbt_max_partition` 

Rendered SQL:
```    select * from (
      


with data as (

    

        -- we want to overwrite the 4 records in the 2020-01-01 partition
        -- with the 2 records below, but add two more in the 2020-01-02 partition
        select 10 as id, cast('2020-01-01' as datetime) as date_time union all
        select 20 as id, cast('2020-01-01' as datetime) as date_time union all
        select 30 as id, cast('2020-01-02' as datetime) as date_time union all
        select 40 as id, cast('2020-01-02' as datetime) as date_time

    

)

select * from data


where date_time >= _dbt_max_partition

    ) as __dbt_sbq
    where false
    limit 0
```

dbt/adapters/bigquery/impl.py
```diff
@@ -72,24 +72,37 @@ class PartitionConfig(dbtClassMixin):
     time_ingestion_partitioning: bool = False
     copy_partitions: bool = False
 
+    def partitions_for_replacement_data_type(self):
```
colin-rogers-dbt, line 75: I think the semantics here are confusing, maybe: `data_type_for_partition`?

colin-rogers-dbt (approved):
LGTM but I'd like to get a second set of eyes on this

github-christophe-oudar:
LGTM but I might not be the most objective person on that PR 😉

PR #492, opened 2023-01-27
BigQuery_tests conversion

McKnight-42:
resolves #

### Description

Convert tests (location_change, hours_to_expiration, and change_partiton) from integration to new functional test format

### Checklist


tests/functional/bigquery/test_bigquery_location_change.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TestBigqueryInvalidLocation(BaseBigQueryLocation):
+
+    @pytest.fixture(scope="class")
+    def profiles_config_update(self, dbt_profile_target):
+        outputs = {"default": dbt_profile_target}
+        outputs["default"]["location"] = _INVALID_LOCATION
+        return {"test": {"outputs": outputs, "target": "default"}}
+
+    def test_bigquery_location_invalid(self, project):
```
McKnight-42, line 36: Curious if we can better this test, the way our config update happens this is technically valid location as its applied during, is there a way to change this part of profile during trest? tried reaching into adapter_response but wasn't having much luck.

tests/functional/test_location_change.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+import os
+from dbt.tests.util import run_dbt_and_capture
```
colin-rogers-dbt, line 3: I think this might be creating an issue with `tests/functional/test_update_column_policy.py` see [test failure](https://github.com/dbt-labs/dbt-bigquery/actions/runs/4048837842/jobs/6966020678#step:7:464). I wonder if we need to set the location back post test?

McKnight-42: will have that triggered good catch was  had noticed that fresh conversion was having a issue at eod

tests/functional/test_location_change.py
```diff
@@ -0,0 +1,38 @@
...
+        return {
+            "model.sql": _MODEL_SQL,
+        }
+
+
+class TestBigqueryValidLocation(BaseBigQueryLocation):
+
+    def test_bigquery_valid_location(self, project):
+        _, stdout = run_dbt_and_capture()
+        assert "Completed successfully" in stdout
```
VersusFacit, line 24: Naive question: Why not just a `dbt run`? What makes this special that it requires a message string check?

tests/functional/test_location_change.py
VersusFacit: Can we use mock? Pytest has a way to in-runtime override for the test itself. Just checkout other tests in dbt core that use this (if it's db-side may not be possible because that state is of course not sandboxed to test runtime). 

I am concerned a failure mid test operation could fail other tests without us knowing why, that is, unless we can sandbox the environment edit.

tests/functional/test_location_change.py
McKnight-42: was playing around with various ways to grab adapter_response wanted to at least be able to check stdout from call to see of model passes or fails as expected (believe second model should fail functionally as we are changing location mid test). a `dbt_run` could be used will tinker a little more.

tests/functional/test_location_change.py
McKnight-42: @VersusFacit any particular test that stood out as a good example to you? happy to play around with it, curious its only the next test that fails

tests/functional/test_location_change.py
VersusFacit: no problem. Thanks for the context

tests/functional/test_location_change.py
VersusFacit: I was thinking of something like this `https://github.com/dbt-labs/dbt-snowflake/blob/471d07e8b09bfdbee0ec35a7d69bf85b8dc946ca/tests/unit/test_snowflake_adapter.py#L367`. Should self contain (by mocking) any env changes since they're not "really" happening but you can _mock_ that they are. 

May not work if you want to do an end to end integration test, but I could see attacking this in a manner to avoid malforming the env -- just an idea

McKnight-42: yeah some of this mocking stuff for our unittest didn't know if it would apply honestly to the pytest will give it a go all for now malforming a profile variable in the directory. actually thought each test formed its own whole little micro project which meant different profiles for each test as well.

VersusFacit: Well what you said is true on a class level but functions share the class and if you change database state, that doesn't get swapped back either unless you tell it to (although I don't like the idea of changing database state without very a bulletproof rollback safety net in place).

tests/functional/test_location_change.py
```diff
@@ -0,0 +1,42 @@
...
+
+class TestBigqueryInvalidLocation(BaseBigQueryLocation):
+
+    @pytest.fixture(scope="class")
+    def profiles_config_update(self, dbt_profile_target):
+        outputs = {"default": dbt_profile_target}
+        outputs["default"]["location"] = _INVALID_LOCATION
+        yield
+        outputs = {"default": dbt_profile_target}
+        outputs["default"]["location"] = _VALID_LOCATION
```
VersusFacit, line 37: Would be amazing if you could:
```
try:
finally:
```
which would guarantee except in the case the code fails so bad it errs out completely that we flip this back. Unsure though if that paradigm plays nicely with `yield` in pytest. Maybe something to Google?

tests/functional/test_changing_partitions.py
```diff
@@ -1,34 +1,88 @@
...
-            self.assertEqual(result.failures, 0)
+            assert result.status == "pass"
+            assert result.skipped == False
+            assert result.failures == 0
+
+class TestBigQueryChangingPartition(BaseBigQueryChangingPartition):
 
-    @use_profile('bigquery')
-    def test_bigquery_add_partition(self):
+    def test_bigquery_add_partition(self, project):
```
VersusFacit, line 85: nit: if you're mucking around these files, might as well remove `bigquery` from all these functions and stuff. I've been nit-ed on this in dbt-core for leaving postgres in places so might as well keep up with the idiom there 🤷‍♀️

tests/functional/test_location_change.py
colin-rogers-dbt: not sure I follow where/why we want try...finally here, can you expand on this a bit?

McKnight-42: I believe the worry is that if this tests fails during run once we can fuse finally to force it to try and reset the variable again back to default to not break the next test

colin-rogers-dbt (approved):

tests/functional/test_location_change.py
colin-rogers-dbt: I believe the `yield` functionality as managed by pytest guarantees that no matter the result of the test associated with the fixture the code after will be executed

tests/functional/test_location_change.py
McKnight-42: also tried using `update_config_file` to change the specific `location` variable mid run which seems closer to the way test originally worked but couldn't get the profiiles.yml to come out correctly.

VersusFacit (approved):

PR #525, opened 2023-02-10
Convert hook integration tests to functional

colin-rogers-dbt:
resolves #

### Description
Happy to be wrong here but it doesn't look like we need to test this here. 

### Checklist


VersusFacit (dismissed):
i sure do love getting rid of redundant code.

jtcohen6 (commented):
Looks like we added these tests way back in https://github.com/dbt-labs/dbt-core/pull/836. I don't think much of what's being tested here is necessary / specific to `dbt-bigquery`.

The one thing worth calling out, though, is that these are defined inside materializations today:
```
{{ run_hooks(pre_hooks) }}
{{ run_hooks(post_hooks) }}
```
Given that adapters can override materializations, it would be totally possible for `dbt-bigquery` (or another adapter) to forget to include those calls to `run_hooks` in their materialization code. Would we have a way of knowing that? Is there another test that would fail if we removed (e.g.) these lines?

https://github.com/dbt-labs/dbt-bigquery/blob/a7af39aef82a16895bd35d4edc97e4dba6d32a7e/dbt/include/bigquery/macros/materializations/table.sql#L12

https://github.com/dbt-labs/dbt-bigquery/blob/a7af39aef82a16895bd35d4edc97e4dba6d32a7e/dbt/include/bigquery/macros/materializations/table.sql#L37

VersusFacit:
Jerco, thanks for trawling through the history here. I did a quick scan of the repo and this seems to be. If we want to adopt a testing philosophy where we create adapter-repo specific tests to create coverage for macro overrides (seems reasonable), I think we should not remove this test, but perhaps have it pull from an adapter zone variant or have a trivial test in this `dbt-bigquery` repo with a comment explaining its presence. 

I've advised a similar policy on other adapter tests (or been encouraged to do the same).

McKnight-42 (approved):

tests/functional/adapter/hooks/test_run_hooks.py
```diff
@@ -0,0 +1,38 @@
...
+    def check_hooks(self, state, project, host):
+        self.get_ctx_vars(state, project)
+
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            # The create and drop table statements here validate that these hooks run
+            # in the same order that they are defined. Drop before create is an error.
+            # Also check that the table does not exist below.
+            "on-run-start": [
```
mikealfare, line 15: Lots of jinja here. Perhaps set these as separate variables like we do with models and schemas?

PR #542, opened 2023-02-16
Allow column policy_tags to be removed by removing them from model

### LINKED ISSUE #349: [CT-1332] [Bug] Removing the policy_tags property from a column does not remove the policy tag(s)

### Current Behavior

Policy tags cannot be removed from columns by removing the policy_tags property.

### Expected Behavior

Removing the `policy_tags` property from a column will remove the policy tags from that column that dbt initially added.

### Steps To Reproduce

As an example, let's say we define the model's corresponding `.yml` file as follows:
With the following definition:

```version: 2

models:
- name: policy_tag_table
  columns:
    - name: field
      policy_tags:
        - 'need_to_know'
```

The model is executed and the `need_to_know` policy tag is added to the `field` column. Now if we decide that we no longer want the `need_to_know` policy tag attached to the `field` column, one would think that removing the `policy_tags` property would remove the policy tags (similar to the behavior of the `description` property).

```version: 2

models:
- name: policy_tag_table
  columns:
    - name: field
```

However, this does not remove the policy tags. Currently, to use dbt to remove the policy tags, an empty list of policy tags needs to be provided instead:

```version: 2

models:
- name: policy_tag_table
  columns:
    - name: field
      policy_tags:
        - 
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS Monterey 12.6
- Python: 3.10.6
- dbt-core: 1.2.1
- dbt-bigquery: 1.2.0
```

### Additional Context

_No response_

garsir:
resolves #349 

### Description

Pull request contains a fix to the adapter implementing the change suggested by [dbeatty10](https://github.com/dbeatty10) in https://github.com/dbt-labs/dbt-bigquery/issues/349. 

An empty list was chosen as the default value as the adapter requires an empty iterable value and this felt cleanest.

A test (test_delete_column_policy.py)  has also been added to capture this behaviour. First creating a table with an column with policy_tag attached before then removing it. Test will pass if said column has a policy_tag of None.

### Checklist

- [ x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x ] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x ] I have run this code in development and it appears to resolve the stated issue
- [x ] This PR includes tests, or tests are not required/relevant for this PR
- [ x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [ x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)


McKnight-42 (approved):
Fantastic work @garsir tested your changes locally and all seems to be working as expected

PR #561, opened 2023-02-25
Ct 2183/setup tools distro info

mikealfare:
### Description

`dbt-bigquery` is failing integration tests on `main` and `1.4.latest` on the Ubuntu py3.11 image due to an issue with `python-distro-info` and `setuptools` as documented in this issue: https://github.com/pypa/setuptools/issues/3772. This resolves the issue by loosening requirements and preventing `setuptools` from being updated beyond `66.0`.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [ ] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #571, opened 2023-02-28
Revert "Ct 2183/setup tools distro info"

mikealfare:
Reverts dbt-labs/dbt-bigquery#561 so that we can implement our approach for dev-requirements.txt in a single backport with these changes


mikealfare:
I cannot revert this PR because it would revert it back to the broken state, which will not pass CI. Instead, there will need to be two backports into `1.4.latest`

PR #606, opened 2023-03-15
Use contract column order in create table

### LINKED ISSUE #579: [BigQuery] Generate order-agnostic constraints ddl when contract: true
Core issue: https://github.com/dbt-labs/dbt-core/issues/7064

Wrap create_table_as user sql in schema-ordered select statement for models with contract: true

```
{% macro <adapter>__create_table_as(temporary, relation, sql) -%}
    ... 
    {% if config.get('contract', False) %}
      {{ get_assert_columns_equivalent(sql) }}
      {{ get_columns_spec_ddl() }}
      {%- set sql = get_select_subquery(sql) %}
    {% endif %}
  as (
    {{ sql }}
  );
{%- endmacro %}
```

BigQuery create_table_as: https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/include/bigquery/macros/adapters.sql#L57

We won't need to do this for view materializations as they don't support constraints.

gshank:
resolves #579

### Description

To support dbt Core changes, use contract column order in create table as.

### Checklist


dbt/include/bigquery/macros/adapters.sql
```diff
@@ -53,8 +53,10 @@
 
     create or replace table {{ relation }}
       {% if config.get('contract', False) %}
-        {{ get_assert_columns_equivalent(sql) }}
+        {% do log('======== has a contract: ' ~ compiled_code) %}
```
mikealfare, line 56: Why do we log for this adapter, but not for the others? I'm fine either way, but we should probably be consistent.

gshank: I think that was leftover from debugging. I'll remove it. There was some confusion about the difference between "compiled_code" and "sql" in the macros (they're both context methods, so people weren't careful about using the passed in string...)

MichelleArk:
Model contracts are new for 1.5, this does not need a backport 👍

PR #639, opened 2023-03-30
ADAP-400: Support materialized views with create/drop functionality

mikealfare:
resolves https://github.com/dbt-labs/dbt-core/issues/6911

### Description

Stub out the materialized view materialization for Redshift. Under the hood this just creates a view for now. But this can be used to verify all of the plumbing.

While creating the materialization, it was observed that other materialization templates existed outside of the `materialization` directory. These were moved from `adapters.sql` into the `materialization` directory in the appropriate file (e.g. `table.sql`). Some of the macros were also refactored to be smaller bits of code and `dbt-postgres` dependencies were eliminated when found.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-redshift/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-redshift/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)


dbt/include/bigquery/macros/materializations/materialized_view.sql
```diff
@@ -0,0 +1,71 @@
...
+
+{% macro bigquery__drop_materialized_view_sql(relation) %}
+    drop materialized view if exists {{ relation }}
+{% endmacro %}
+
+{% macro bigquery__get_replace_materialized_view_as_sql(
+    relation,
+    sql,
+    existing_relation,
+    backup_relation,
```
McKnight-42, line 13: `backup_relation` and `intermediate_relation` don't appear to be used anywhere else in `dbt-bigquery` curious if we will have to make default versions of most the chunk parts of matierliazation we define in `adap-2` to be able to handle the changes `base` adapters will have as they don't deal with a lot of pre-built SQL features nateively

McKnight-42: can we just add `temp_relation` to the input fields as needed? do we also need to do that to dbt-core default versions?

tests/functional/adapter/materialized_view_tests/test_materialized_view.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        assert_model_exists_and_is_correct_type(project, "base_table", RelationType.Table)
+
+    def test_relation_is_materialized_view_when_rerun(self, project):
+        run_model("base_materialized_view")
+        assert_model_exists_and_is_correct_type(
+            project, "base_materialized_view", RelationType.MaterializedView
+        )
+
+    # def test_relation_is_materialized_view_on_full_refresh(self, project):
```
mikealfare, line 30: I use `@pytest.mark.skip("<message for why we're skipping>")` in these scenarios. It's more readable because you get the formatting back, tests that should be skipped will still likely get updated since the syntax checker will hit it, and there's a built in message as to why these tests are skipped.

dbt/include/bigquery/macros/materializations/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,38 @@
+{% materialization materialized_view, adapter='bigquery' %}
```
mikealfare, line 1: Do we have to rewrite the materialization? I think we should be able to use the default one in `dbt-core`.

dbt/include/bigquery/macros/materializations/materialized_view/materialized_view.sql
```diff
@@ -0,0 +1,38 @@
...
+        {% endif %}
+
+    {{ run_hooks(post_hooks, inside_transaction=False) }}
+
+    {{ return({'relations': [target_relation]}) }}
+
+{% endmaterialization %}
+
+
+{% materialized_view_get_build_sql(existing_relation, target_relation, backup_relation, intermediate_relation) %}
```
mikealfare, line 24: I also think this could be replaced by choosing appropriate macros in `macros/relations/materialized_view/*.sql`. For example, the alter macro should just call replace. Also, since we're returning an empty changeset, we shouldn't get past the call to refresh.

dbt/include/bigquery/macros/relations/drop.sql
```diff
@@ -3,3 +3,12 @@
     drop {{ relation.type }} if exists {{ relation }}
   {%- endcall %}
 {% endmacro %}
+
+
+{% macro bigquery__drop_relation_sql(relation) -%}
```
mikealfare, line 8: Was this written specifically for this work? Or does something else call `bigquery__drop_relation_sql` or `drop_relation_sql`? If it's the former, then we should rework the dependent macros to use the new `dbt-core` versions. Ideally we don't have this file at all, but I think we implemented a generic `drop_relation` macro for some reason (maybe to remove the cascade keyword?).

dbt/include/bigquery/macros/relations/materialized_view/_replace.sql
mikealfare, line 1: I put a leading underscore on this because we shouldn't have a replace macro in here. I didn't go through and check where it's used. Wherever it's used, it should be replaced with the replace macro in `dbt-core`, `get_replace_sql`.

dbt/include/bigquery/macros/relations/materialized_view/alter.sql
```diff
@@ -0,0 +1,16 @@
+{% macro bigquery__get_alter_materialized_view_as_sql(
+    relation,
+    configuration_changes,
+    sql,
+    existing_relation,
+    intermediate_relation
+) %}
+    get_replace_materialized_view_as_sql(
```
mikealfare, line 8: As noted above, replace this with `get_replace_sql`. This macro won't need intermediate_relation or backup_relation. What's temp_relation?

dbt/include/bigquery/macros/relations/materialized_view/refresh.sql
```diff
@@ -0,0 +1,10 @@
+{% macro bigquery__refresh_materialized_view(relation) %}
+    get_replace_materialized_view_as_sql(
```
mikealfare, line 2: I probably mentioned to do this initially, but let's replace this with the correct ddl.

tests/functional/adapter/materialized_view_tests/test_materialized_view.py
mikealfare, line 1: I haven't cross referenced this against what we run for MV now; a lot has changed since we started. I would just check it again, including the fixtures file.

dbt/adapters/bigquery/relation.py
```diff
@@ -16,6 +17,14 @@
 class BigQueryRelation(BaseRelation):
     quote_character: str = "`"
     location: Optional[str] = None
+    # why do we need to use default_factory here but we can assign it directly in dbt-postgres?
```
mikealfare, line 20: Does anyone know why I needed to use field(default_factory...) here when `dbt-postgres` allows me to register this with a simple list? Removing this causes unit tests to fail due to mashumaro errors for using a mutable default.

mikealfare: Figured it out, we don't test for this in `dbt-postgres`. This does not work with `.validate()` unless we use `default_factory`.

McKnight-42 (approved):
Looks like this is at the end state me and Mike worked toward two weeks ago and put in line for the mergers of initial prs on remaining work.

PR #653, opened 2023-04-08
add steps to drop existing tables on --full-refresh

### LINKED ISSUE #609: [ADAP-380] [Bug] --full-refresh doesn't work for python models

### Current Behavior

When writing a python dbt model backed by BigQuery, if I change the output model schema (for example by changing the format of the returned DataFrame), then the DataProc job fails with:

1) [Guice/ErrorInCustomProvider]: IllegalArgumentException: BigQueryConnectorException$InvalidSchemaException: Destination table's schema is not compatible with dataframe's schema
  at BigQueryDataSourceWriterModule.provideDirectDataSourceWriterContext(BigQueryDataSourceWriterModule.java:60)
  while locating BigQueryDirectDataSourceWriterContext

This still occurs if I run dbt with --full-refresh

### Expected Behavior

When using full-refresh, I would expect the output table to be dropped before executing the DataProc job, so this kind of error would not happen. 

### Steps To Reproduce

1. Write dbt python job that runs successfully and creates the resulting output table.
2. Change the job to have a different schema (e.g. type change on a column)
3. Run dbt run --full-refresh

### Relevant log output

_No response_

### Environment

```markdown
- OS:Ubuntu 20.04
- Python: 3.10.9
- dbt-core: 1.4.1
- dbt-bigquery: 1.4.2
```

### Additional Context

_No response_

VersusFacit:
resolves #609 

### Description

I'm confident I've got a working fix here; a little less confident on user experience. 

This is also my first time working with all of these systems involved with python models; interesting stuff. Lots of surface area to find a small fix.

#### 1. the root of the problem
When schema changes are performed inside a model and the target relation already exists (albeit in a now out-of-date form), the bigquery spark overwrite strategy fails ([I checked, there are no built-in alternatives](https://spark.apache.org/docs/1.4.0/api/java/org/apache/spark/sql/SaveMode.html)). 

#### 2. my proposed solution
Therefore, a step is needed to clear out any existing, corresponding target relation before attempting to build a model. I first attempted to do this inside of the Python model, but that only clears the model in the spark namespace. We must pre-clear the model *only* in the SQL namespace. 

#### 3. one open question
@Fleid: I predict midflight schema changes are going to occur when a user is intentionally changing the definition of a model, which I don't think would be an everyday occurrence. We can therefore confine this behavior to the `--full-refresh` flag which empowers the user to decide when to invoke this behavior. This is only needed in this schema change scenario because of a quirk of `save`, a dataframe function. 

I don't love the idea of overloading this flag further, but I also think it is relatively intuitive and better to make this opt in; can't think of a better differentiator. Plus, we already say in [docs](https://docs.getdbt.com/reference/resource-configs/full_refresh) that models generically are affected by this flag.

Do you let me know if you think we should pursue a different user experience

### Testing regiment

```py 
import pandas
def model(dbt, spark):
    dbt.config(
        materialized='table',
    )
    data = [[1,2]] * 10
    return spark.createDataFrame(data, schema=['test', 'test2'])
```
First, run this. Second, change the model to this:
```py
import pandas
def model(dbt, spark):
    dbt.config(
        materialized='table',
    )
    data = [[1,2]] * 10
    return spark.createDataFrame(data, schema=['test1', 'test3'])
```
When you re-run this without `--full-refresh` or my changes, the error discussed in the issue would be triggered. Here's an example:

image

image

But with the new behavior of `dbt run -m <py model> --full-refresh`, things look good.

image

image

image

### Checklist

### DIFF
--- a/dbt/include/bigquery/macros/adapters.sql
+++ b/dbt/include/bigquery/macros/adapters.sql
@@ -73,6 +73,12 @@
     TODO: Deep dive into spark sessions to see if we can reuse a single session for an entire
     dbt invocation.
      --#}
+
+    {#-- when a user wants to change the schema of a relation, first drop the table in the big query dataset --#}
+    {%- set old_relation = adapter.get_relation(database=relation.database, schema=relation.schema, identifier=relation.identifier) -%}
+    {%- if (old_relation.is_table and (should_full_refresh())) -%}
+      {% do adapter.drop_relation(relation) %}
+    {%- endif -%}
     {{ py_write_table(compiled_code=compiled_code, target_relation=relation.quote(database=False, schema=False, identifier=False)) }}
   {%- else -%}
     {% do exceptions.raise_compiler_error("bigquery__create_table_as macro didn't get supported language, it got %s" % language) %}
--- a/dbt/include/bigquery/macros/adapters.sql
+++ b/dbt/include/bigquery/macros/adapters.sql
@@ -74,7 +74,7 @@
     dbt invocation.
      --#}
 
-    {#-- when a user wants to change the schema of a relation, first drop the table in the big query dataset --#}
+    {#-- when a user wants to change the schema of an existing relation, they must intentionally drop the table in the dataset --#}
     {%- set old_relation = adapter.get_relation(database=relation.database, schema=relation.schema, identifier=relation.identifier) -%}
     {%- if (old_relation.is_table and (should_full_refresh())) -%}
       {% do adapter.drop_relation(relation) %}
--- a/.changes/unreleased/Fixes-20230408-035117.yaml
+++ b/.changes/unreleased/Fixes-20230408-035117.yaml
@@ -1,0 +1,7 @@
+kind: Fixes
+body: ' add full refresh capabilities to tabular bigquery python models to accommodate
+  schema changes'
+time: 2023-04-08T03:51:17.167349-07:00
+custom:
+  Author: versusfacit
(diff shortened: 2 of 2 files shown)


VersusFacit:
Also, I noticed tests exist that check this behavior, but currently we skip them. I tried un-skipping them and ran into some bugs. Apparently the note left that our system is fragile holds true. 

For now I've made you've been simulating that test to check that my changes to what I expected to do

Fleid:
In terms of user experience, I like that we're pushing the new behavior into `--full-refresh`. 
Good thinking and thanks a lot Mila!

dbt/include/bigquery/macros/adapters.sql
```diff
@@ -73,6 +73,12 @@
     TODO: Deep dive into spark sessions to see if we can reuse a single session for an entire
     dbt invocation.
      --#}
+
+    {#-- when a user wants to change the schema of an existing relation, they must intentionally drop the table in the dataset --#}
```
mikealfare, line 77: Do we specifically only want to do this when the old relation is a table? Or do we always want to drop the old relation if the names collide (e.g. a view is now a table)? If it's the latter, I think this can be boiled down to `{{ drop_relation_if_exists(relation) }}`

mikealfare: Oh interesting. Ok, makes sense. Thanks for the context.

mikealfare (approved):

PR #688, opened 2023-04-28
Fixes adding policy tags when a struct is defined in the yml

dgreen161:
resolves #687

### Description

This resolves the issue where policy tags attempting to be applied to structs (records) in BigQuery. These can only be applied to leaf nodes as described in the docs [here](https://cloud.google.com/bigquery/docs/column-level-security#struct_or_record)

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

Issue #687, opened 2023-04-28
[ADAP-503] [Regression] Persisted Metadata does not update on BigQuery tables


patricker:
### Is this a new bug in dbt-core?

- [X] I believe this is a new bug in dbt-core
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

After upgrading to dbt 1.5.0 column metadata does not update in BigQuery and the model build shows as failed.

### Expected Behavior

metadata writes to the tables without issue in BigQuery.

### Steps To Reproduce

1. with dbt 1.5.0, writing to BigQuery
2. Update your dbt_project:
    +persist_docs:
      relation: true
      columns: true
3. Run a model with column level documentation
4. Error received after model is fully built, during the docs persist phase.

### Relevant log output

```shell
Error: google.api_core.exceptions.BadRequest: 400 PATCH https://bigquery.googleapis.com/bigquery/v2/projects/{table_path}?prettyPrint=false: Policy tags can only be attached to a leaf field.
```

### Environment

```markdown
- OS:debian
- Python:3.9.10
- dbt:1.5.0
```

### Which database adapter are you using with dbt?

bigquery

### Additional Context

I rolled back to 1.4.6 and the issue went away.

dbeatty10:
Thanks for reporting this @patricker and @steve-marmalade.

I was able to reproduce this when using [`policy_tags`](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#policy-tags) and [`persist_docs`](https://docs.getdbt.com/reference/resource-configs/persist_docs).

dbeatty10:
### Simple reprex

`dbt_project.yml`
```yaml
name: "my_dbt_project"
version: "1.0.0"
config-version: 2
profile: "sandcastle-bigquery"

models:
  my_dbt_project:
    +persist_docs:
      relation: true
      columns: true
```

`models/_models.yml`
```yaml
# version isn't required for dbt-core 1.5, but it is for 1.4
version: 2

models:
  - name: my_model
    columns:
      - name: id
        description: Primary key
        policy_tags:
          # Example pattern:
          # - 'projects/<gcp-project>/locations/<location>/taxonomies/<organization>/policyTags/<tag>'
          - projects/dbt-test-env/locations/us/taxonomies/5785568062805976401/policyTags/135489647357012267
```

`models/my_model.sql`
```sql
{{ config(materialized='table') }}

select 1 as id
```

[...]

PR #607, opened 2023-03-15
Remove checks on non-leaf-field tags since they no longer can exist.

VersusFacit:
resolves #605 

### Description

Apparently -- cannot yet find evidence of this -- bigquery no longer accepts that parent models contain populate fields; only their leaf nodes can. I've asked in the DBT Slack if anyone has information as a normal debugging routine has turned up nothing. I saw a suspect google decided in this newest version of the API to close a loophole or perhaps disbar a previously deprecated feature.

This one will require backports!

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR
- [x] I have run `changie new` to [create a changelog entry](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md#Adding-CHANGELOG-Entry)

PR #694, opened 2023-05-03
Revert PR 607

VersusFacit:
# Dr. Strangebug: How I Learned to Stop Worrying and Love the Code

Prompted by issue #687 and PR #688 

### Description

Stemming from a lack of documentation, I didn't realize the patch 400 error was happening because policy tags were being applied (but only an empty list of them!) to non leaf fields, namely the inner struct (which itself has an inner struct with leaf fields). My PR #607  was misguided as a result.

#### What's really going on

My current working understanding is that a leaf field is something tied to a STRUCT(). These can have inner (cf. nested) STRUCTS, but these are not leaf fields. Only true columns are leaf fields. Thus, if the data type happens to be a record (the internal type of a STRUCT), his policy tag mechanism errors out with the 400 response.

In pr #542 , this bug was introduced indirectly [at this line](https://github.com/dbt-labs/dbt-bigquery/pull/542/files#diff-96ff53dca401b9ca48009a58b6c77181552b72b2ad6867c37ae2f22f75a5bd38L647) -- we no longer forbid 'empty' policy tag PATCH-es. This happened even in cases with no policy tags because of how `impl.py` has at least an empty list applied to columns (this part is still obscure to me because it happens in BQ's code, not ours -- **_why would an empty list attempt a patch on a model?_**).

#### The root cause

Turns out [we didn't run adapter tests](https://github.com/dbt-labs/dbt-bigquery/pull/688/files) on PR #542 . This is a reminder we must add all test tags to community PRs before merging.

### Checklist

- [X] I have read [the contributing guide](https://github.com/dbt-labs/dbt-bigquery/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [X] I have signed the [CLA](https://docs.getdbt.com/docs/contributor-license-agreements)
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] I have [opened an issue to add/update docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose), or docs changes are not required/relevant for this PR

### DIFF
--- a/tests/functional/adapter/test_persist_docs.py
+++ b/tests/functional/adapter/test_persist_docs.py
@@ -36,10 +36,18 @@ version: 2
 models:
   - name: table_model_nested
     columns:
+      - name: level_1
+        description: level_1 column description
+      - name: level_1.level_2
+        description: level_2 column description
       - name: level_1.level_2.level_3_a
         description: level_3 column description
   - name: view_model_nested
     columns:
+      - name: level_1
+        description: level_1 column description
+      - name: level_1.level_2
+        description: level_2 column description
       - name: level_1.level_2.level_3_a
         description: level_3 column description
 """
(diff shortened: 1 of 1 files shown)


McKnight-42 (approved):
Really great work! love the detail in backtracking to source and the lay out of this

nssalian (approved):
Thanks for the revert and the detailed explanation.

Issue #734, opened 2023-05-22
[ADAP-569] [Enhancement] only wait for Dataproc job to finish, not for the Dataproc serverless cluster to spin down


wazi55:
### Is this a new bug in dbt-bigquery?

- [X] I believe this is a new bug in dbt-bigquery
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When working with DBT with Dataproc, the current behavior is DBT waits for the batch job complete the torn down of the cluster [here](https://github.com/dbt-labs/dbt-bigquery/blob/7bc2af0cef1e42b625cee6d996b48cb91dc2e49c/dbt/adapters/bigquery/python_submissions.py#L127), however, that adds an additional 1 or 2 minutes on the computation time.  Instead of waiting for the job finish, the code should just wait until the job is finished by polling using the get_batch() method to check whether the job has succeeded or not. 

### Expected Behavior

The expected behavior is, DBT should continue on with the SQL models after the spark code succeed instead of waiting for the cluster to be torn down. 

### Steps To Reproduce

```
    def _submit_dataproc_job(self) -> dataproc_v1.types.jobs.Job:
        batch = self._configure_batch()
        parent = f"projects/{self.credential.execution_project}/locations/{self.credential.dataproc_region}"

        request = dataproc_v1.CreateBatchRequest(
            parent=parent,
            batch=batch,
        )
        # make the request
        operation = self.job_client.create_batch(request=request)  # type: ignore
        # this takes quite a while, waiting on GCP response to resolve
        # (not a google-api-core issue, more likely a dataproc serverless issue)
        response = operation.result(retry=self.retry)
        return response
```

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-bigquery:
```

dbeatty10:
Thanks for reporting this @wazi55!

Are you interested in contributing a pull request for this, by any chance?

dbeatty10:
I re-labeled this as an `enhancement` since I don't perceive there to be an error, flaw, failure or fault here -- this is more of an efficiency / optimization thing.

dataders:
@wazi55 thanks so much for opening! This was on my to-do list to write up after an email thread last month with Big Query engineers.

[...]

PR #766, opened 2023-06-09
Pass python model timeout to polling instead of retry

### LINKED ISSUE #577: [ADAP-346] job_execution_timeout_seconds config variable not used when running python models on dataproc. Python models timeout after 900 seconds

### Current Behavior

We are currently running 3 python models in our DAG. Both locally and in Airflow (once a month). One of our python models use a lot of data and takes some time to complete (normally 20-30 min). When we were using dbt-core 1.3.0 in production there were no issues until the dataproc bug was introduced [bug](https://github.com/dbt-labs/dbt-bigquery/issues/531). This forced us to upgrade dbt-bigquery and dbt-core. 

DBT Python models running in dataproc time out after 15 minutes even though we have set 
`job_execution_timeout_seconds: 7200` - both in the repo and in our ~/.dbt folder. We have also tried adding it into the model-specific .yml file but without success.

Running dbt --debug gave us where the 900 default timeout is coming from. The google package python-core-api sets the default timeout:

python3.9/site-packages/google/api_core/future/polling.py:
`# Default polling configuration
DEFAULT_POLLING = retries.Retry(
    predicate=POLLING_PREDICATE,
    initial=1.0,  # seconds
    maximum=20.0,  # seconds
    multiplier=1.5,
    timeout=900,  # seconds
)`

My question is: why isn't our timeout being used for python models? And what is the best way to fix this? Right now we are editing the polling.py file to timeout after 7200 seconds which works but isn't feasible long-term.

### Expected Behavior

dbt python models running in dataproc use the `job_execution_timeout_seconds: 7200` variable set in our profiles.yml

### Steps To Reproduce

1. dbt-bigquery 1.3.1 dbt-core 1.3.1 (dbt-bigquery 1.4.x has bugs related to incremental models being materialized as tables)
2. Set `job_execution_timeout_seconds: 7200`
3. Run a python model which takes longer than 15 min to complete

### Relevant log output

_No response_

### Environment

```markdown
- OS: Mac OS Ventura 13.0
- Python: 3.9.6
- dbt-core: 1.3.1
- dbt-bigquery: 1.3.1
```

### Additional Context

_No response_

colin-rogers-dbt:
resolves #577 

### Description
GCP async interface offers two configurations that both accept the same type: `retry` and `polling`. The difference is that `retry` configures each individual polling request while `polling` configures the long running operation (i.e. how long to poll for and how frequently to poll). We have been setting `retry` and not `polling` resulting in unexpected behavior in job configuration. 

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/python_submissions.py
+++ b/dbt/adapters/bigquery/python_submissions.py
@@ -45,7 +45,9 @@ class BaseDataProcHelper(PythonJobHelper):
         self.timeout = self.parsed_model["config"].get(
             "timeout", self.credential.job_execution_timeout_seconds or 60 * 60 * 24
         )
-        self.retry = retry.Retry(predicate=POLLING_PREDICATE, maximum=10.0, timeout=self.timeout)
+        self.result_polling_policy = retry.Retry(
+            predicate=POLLING_PREDICATE, maximum=10.0, timeout=self.timeout
+        )
         self.client_options = ClientOptions(
             api_endpoint="{}-dataproc.googleapis.com:443".format(self.credential.dataproc_region)
         )
@@ -100,7 +102,7 @@ class ClusterDataprocHelper(BaseDataProcHelper):
                 "job": job,
             }
         )
-        response = operation.result(polling=self.retry)
+        response = operation.result(polling=self.result_polling_policy)
         # check if job failed
         if response.status.state == 6:
             raise ValueError(response.status.details)
@@ -125,7 +127,7 @@ class ServerlessDataProcHelper(BaseDataProcHelper):
         operation = self.job_client.create_batch(request=request)  # type: ignore
         # this takes quite a while, waiting on GCP response to resolve
         # (not a google-api-core issue, more likely a dataproc serverless issue)
-        response = operation.result(polling=self.retry)
+        response = operation.result(polling=self.result_polling_policy)
         return response
         # there might be useful results here that we can parse and return
         # Dataproc job output is saved to the Cloud Storage bucket
--- a/dbt/adapters/bigquery/python_submissions.py
+++ b/dbt/adapters/bigquery/python_submissions.py
@@ -1,6 +1,8 @@
 from typing import Dict, Union
 
 from dbt.adapters.base import PythonJobHelper
+from google.api_core.future.polling import POLLING_PREDICATE
+
 from dbt.adapters.bigquery import BigQueryConnectionManager, BigQueryCredentials
 from dbt.adapters.bigquery.connections import DataprocBatchConfig
 from google.api_core import retry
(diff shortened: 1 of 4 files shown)


dbt/adapters/bigquery/python_submissions.py
```diff
@@ -43,7 +45,7 @@ def __init__(self, parsed_model: Dict, credential: BigQueryCredentials) -> None:
         self.timeout = self.parsed_model["config"].get(
             "timeout", self.credential.job_execution_timeout_seconds or 60 * 60 * 24
         )
-        self.retry = retry.Retry(maximum=10.0, deadline=self.timeout)
+        self.retry = retry.Retry(predicate=POLLING_PREDICATE, maximum=10.0, timeout=self.timeout)
```
github-christophe-oudar, line 48: According to https://googleapis.dev/python/google-api-core/latest/_modules/google/api_core/future/polling.html
`predicate=POLLING_PREDICATE,` is already a default so you can skip it

```suggestion
        self.retry = retry.Retry(maximum=10.0, timeout=self.timeout)
```

github-christophe-oudar (approved):
nitpicking, LGTM

dbt/adapters/bigquery/python_submissions.py
colin-rogers-dbt: My read on this is that if we supply a custom `Retry` to polling (i.e. that replaces the `DEFAULT_POLLING` retry) it won't necessarily add the default polling predicate

dbt/adapters/bigquery/python_submissions.py
```diff
@@ -98,7 +100,7 @@ def _submit_dataproc_job(self) -> dataproc_v1.types.jobs.Job:
                 "job": job,
             }
         )
-        response = operation.result(retry=self.retry)
+        response = operation.result(polling=self.retry)
```
ChenyuLInx, line 103: Should we consider rename `self.retry` to some thing different like `self.poll_policy` or `self.get_result_policy`.

ChenyuLInx (approved):
Thanks for adding this!!

Completely out of scope for this issue but something slightly related just for providing context, when submitting python job using `serverless(batch job instead of a pre-created cluster)`, we currently actually waiting for the serverless Dataproc cluster fully destroyed before reporting model finish. It would be helpful to have dbt continue on executing when the job itself is done. Related issue at https://github.com/dbt-labs/dbt-bigquery/issues/734

dbt/adapters/bigquery/python_submissions.py
colin-rogers-dbt: good call, I think `result_polling_policy` is the most explicit

dbt/adapters/bigquery/python_submissions.py
github-christophe-oudar: Yeah I guess you're right, the `DEFAULT_POLLING` is not used anymore then and [the retry constructor doesn't have a predicate by default](https://googleapis.dev/python/google-api-core/latest/retry.html)

McKnight-42 (approved):

PR #803, opened 2023-06-29
add clean_up method for store_test_failures tests to stop hanging art…

McKnight-42:
…ifacts in bigquery

resolves # hanging artifacts left during test runs due to change in schema name not caught by default teardown method

### Description

would look to what branches we need to backport

### Checklist

### DIFF
--- a/tests/functional/adapter/store_test_failures_tests/test_store_test_failures.py
+++ b/tests/functional/adapter/store_test_failures_tests/test_store_test_failures.py
@@ -1,9 +1,22 @@
+import pytest
 from dbt.tests.adapter.store_test_failures_tests.test_store_test_failures import (
     StoreTestFailuresBase,
 )
 
 
+TEST_AUDIT_SCHEMA_SUFFIX = "dbt_test__aud"
+
+
 class TestBigQueryStoreTestFailures(StoreTestFailuresBase):
-    def test_store_and_assert(self, project):
+    @pytest.fixture(scope="function")
+    def clean_up(self, project):
+        yield
+        relation = project.adapter.Relation.create(
+            database=project.database, schema=f"{project.test_schema}_{TEST_AUDIT_SCHEMA_SUFFIX}"
+        )
+
+        project.adapter.drop_schema(relation)
+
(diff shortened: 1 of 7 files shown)


tests/functional/adapter/store_test_failures_tests/test_store_test_failures.py
```diff
@@ -1,9 +1,22 @@
...
 )
 
 
+TEST_AUDIT_SCHEMA_SUFFIX = "dbt_test__aud"
+
+
 class TestBigQueryStoreTestFailures(StoreTestFailuresBase):
-    def test_store_and_assert(self, project):
+    @pytest.fixture(scope="function")
+    def clean_up(self, project):
```
mikealfare, line 12: This looks like a teardown method. In my experience, such a method tends to get called `teardown_method` ("method" not because this is a method, but because the scope is "function") and runs automatically.

With that in mind, would you mind if we rename it, add `autouse=True` in the fixture decorator, and remove `clean_up` from the signature of `test_store_and_assert`? It would all behave the same (unless `StoreTestFailuresBase` has additional tests), but it would follow conventions that I've seen.

tests/functional/adapter/store_test_failures_tests/test_store_test_failures.py
```diff
@@ -1,9 +1,22 @@
...
     StoreTestFailuresBase,
 )
 
 
+TEST_AUDIT_SCHEMA_SUFFIX = "dbt_test__aud"
+
+
 class TestBigQueryStoreTestFailures(StoreTestFailuresBase):
+    @pytest.fixture(autouse=True)
+    def teardown_method(self, project):
```
McKnight-42, line 12: should this be something we go change in all tests that have a `clean_up` method instaed

mikealfare: I don't think we need to actively change it, but if we're adding new items or updating something, then I'd change it as part of that. Also, you still need `scope="function"`, so it would be `@pytest.fixture(scope="function", autouse=True)`

mikealfare (approved):

PR #866, opened 2023-08-07
Add an option to use INFORMATION_SCHEMA for partition info retrieval

### LINKED ISSUE #867: [ADAP-781] [Feature] Add an option to use INFORMATION_SCHEMA for partition info retrieval

### Describe the feature

So far there are few places where dbt-bigquery access partitions information:
- In incremental materialization, `_dbt_max_partition` is declared (when used in a model) and stores the max partition
- In `get_partitions_metadata` macro that returns the partitions information for input model
- In incremental `insert_overwrite` materialization and before the MERGE statement with dynamic partitions to store them in `dbt_partitions_for_replacement` and use it in the predicate
- In incremental `insert_overwrite` materialization and before the copy partitions statement with dynamic partitions to iterate on the partitions to replace

Allowing to use `INFORMATION_SCHEMA` data would reduce the amount of data scanned.
More details are available on https://cloud.google.com/bigquery/docs/information-schema-partitions.

As this table is still in Preview, it's likely safer to provide as an alternative to current method.

To access that new method, adding `partition_information` settings to `partition_by` config block so that without value the default is current behavior or with explicit value "model" or use information schema with value "information_schema" to use the new approach.

This issue is similar to https://github.com/dbt-labs/dbt-bigquery/issues/286 but intend to rely on existing macros and solutions as well replacing under the hood accesses to improve overall performances.

We could try to set the config at connection level (profiles.yml) but doing so would change the API for existing macros and wouldn't be as convenient as accessing model config block.

### Who will this benefit?

Anyone using incremental strategy with fairly large partitions as the current approach can be costly and and slightly slow compared to the metadata approach.

yes

### Anything else?

_No response_

github-christophe-oudar:
resolves #867

### Problem

Current approach to retrieve partition information is expensive and slow.

### Solution

Providing a way to use `INFORMATION_SCHEMA` metadata would speed up the partition info retrieval as well for a fraction of the cost (almost constant to 10 MB)

### Checklist

### DIFF
--- a/dbt/include/bigquery/macros/etc.sql
+++ b/dbt/include/bigquery/macros/etc.sql
@@ -6,6 +6,13 @@
   {% do adapter.grant_access_to(entity, entity_type, role, grant_target_dict) %}
 {% endmacro %}
 
+{#
+  This macro returns the partition metadata for provided table.
+  The expected input is a table object (ie through a `source` or `ref`).
+  The output contains the result from partitions information for your input table.
+  The details of the retrieved columns can be found on https://cloud.google.com/bigquery/docs/managing-partitioned-tables
+  It will leverage the INFORMATION_SCHEMA.PARTITIONS table.
+#}
 {%- macro get_partitions_metadata(table) -%}
   {%- if execute -%}
     {%- set res = adapter.get_partitions_metadata(table) -%}
--- a/.changes/unreleased/Features-20230807-235539.yaml
+++ b/.changes/unreleased/Features-20230807-235539.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add an option to use INFORMATION_SCHEMA for partition info retrieval
+time: 2023-08-07T23:55:39.31409+02:00
+custom:
+  Author: Kayrnt
+  Issue: "867"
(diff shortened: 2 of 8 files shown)


dbt/include/bigquery/macros/etc.sql
```diff
@@ -6,6 +6,13 @@
...
 {% endmacro %}
 
+{#
+  This macro returns the partition metadata for provided table.
+  The expected input is a table object (ie through a `source` or `ref`).
+  The output contains the result from partitions information for your input table.
+  The details of the retrieved columns can be found on https://cloud.google.com/bigquery/docs/managing-partitioned-tables
+  It will leverage the INFORMATION_SCHEMA.PARTITIONS table.
+#}
 {%- macro get_partitions_metadata(table) -%}
```
github-christophe-oudar, line 16: I was hoping to add an option to keep the previous approach but I can't since the prototype of `adapter.get_partitions_metadata` is set in dbt-core and I don't think we want to leak the BQ specific option.
I think that, overall, it's acceptable.

nathaniel-may:
Hi @github-christophe-oudar, we appreciate the effort you took to implement this optimization and we thank you for your patience. I've made sure this is properly filed in our queue of work but it may still take some time for us to give it a proper review.

HbirdJ:
I don't think this should be closed, there's a long history of requests for review here and in the dbt bq slack channel.

github-christophe-oudar:
https://github.com/dbt-labs/dbt-bigquery/issues/867 got closed but I'm not sure it should be canceled.
If @amychen1776 referred to microbatch feature, I think we should proceed as it's not the same need.
I would probably prefer merging https://github.com/dbt-labs/dbt-bigquery/pull/1351 before because this one needs a rebase anyway and I refactored a bit in that PR.

PR #868, opened 2023-08-08
Remove walrus operator to make 1.5 py37 compatible

mikealfare:
### Problem

We backported a change to v1.5 that used the walrus operator, which was introduces in py38. While our goal is to drop support for py37 going forward, this is a small change that would keep v1.5 py37 compatible for a bit longer.

### Solution

Replace the walrus operator with an equivalent two lines.

### Checklist


mikealfare:
I converted this to a draft since we might choose to remove testing for py37 as an alternative.

mikealfare:
We are instead choosing to remove py37 testing. This is not needed.

PR #901, opened 2023-08-28
[ADAP-851] remove some connection fields that may be sensitive user information …

### LINKED ISSUE #900: [ADAP-851] [Regression] debug outputs sensitive information

### Current Behavior

related: https://github.com/dbt-labs/dbt-snowflake/issues/754

this change was introduced in #754 (core PR: https://github.com/dbt-labs/dbt-core/pull/7741) and landed in version 1.6.0

`dbt debug` will return the following profile target attributes if the user has them in their profile:
- `token`
- `client_secret`
- `keyfile_json`

The solution is to modify `BigQueryCredentials._connection_keys()` (see below)

https://github.com/dbt-labs/dbt-bigquery/blob/d7fb235a42780841c82ea5b01a647b3416070afe/dbt/adapters/bigquery/connections.py#L174-L194

#### `._connection_keys()` context

our [Building a new Adapter: editing the connection manager](https://docs.getdbt.com/guides/dbt-ecosystem/adapter-development/3-building-a-new-adapter#editing-the-connection-manager) stipulates
> the Credentials’ _connection_keys method […] will return the keys that should be displayed in the output of the dbt debug command. As a general rule, it’s good to return all the arguments used in connecting to the actual database except the password (even optional arguments).

#### longer-term solution?
potential 2.0 work item would be to invert that current pattern and have a place to define sensitive keys that should not be printed to stdout, and by default include all others?

`dbt debug` will not print sensitive connection information to stdout

### Steps To Reproduce

1. use dbt-bigquery version 1.6.0.
2. have a sensitive-field that isn't `password` defined.
3. execute `dbt debug`

McKnight-42:
…that we don't want showing up in dbt debug commands

resolves #900 

### Problem
potentially user sensitive connection details can be added to logs as stdout via dbt debug command

### Solution
remove those creds from `_connection_keys` method so as they are not printed out.

### Checklist


dbt/adapters/bigquery/connections.py
```diff
@@ -186,12 +186,9 @@ def _connection_keys(self):
             "job_creation_timeout_seconds",
```
MichelleArk, line 186: can't comment further up on linen 174 - but it'd probably be worth a comment there [linking to the docs on this method](https://docs.getdbt.com/guides/dbt-ecosystem/adapter-development/3-building-a-new-adapter#editing-the-connection-manager), and indicating that any elements in the returned set are included in debug stdout so secrets should be omitted.

colin-rogers-dbt (approved):

PR #902, opened 2023-08-28
Timeout BQ queries 

### LINKED ISSUE #231: [CT-881] refactor retries
We currently have a non-trivial amount of code to support our homegrown retry strategy via `_retry_and_handle`. This in inferior to using Google's provided and configurable `Retry()` decorator.

This is an extension to the work started in #230.

Goals:
- Make sure all network calls have the `Retry()` decorator.
- If the user provides configuration for the job_* retry/timeout options, plumb them into the Retry() decorator in the appropriate ways.
- Delete homegrown retry code

Context:
It is reasonable that a user would want a very quick timeout if a job fails to create/submit, and a much much _much_ longer timeout for a job that's actually running (e.g. models that take >1 hr to run).

colin-rogers-dbt:
Time out and cancel a model execution if it exceeds the provided number of seconds. 

Currently the bigquery python client does not provide us with a mechanism to specify a timeout on the database side. As a workaround until they do we can do the next best thing: timeout the execution, cancel the query and raise an exception. 

Note we will only use this new logic if a user specifies the `job_execution_timeout` parameter in their profile. If it is not set we continue using the same behavior as today.

resolves #231

### Checklist


github-christophe-oudar:
It's on my wishlist for dbt features!
I though about parsing dbt logs to cancel BQ jobs (parsing the URLs) at scheduler level but if it can support canceling queries on SIGTERM signals, it would be great 🙌 
Thank you for working on this!

colin-rogers-dbt:
> if it can support canceling queries on SIGTERM signals, it would be great
@github-christophe-oudar don't know if we'll tackle that in this pr as I think this calls for a consistent approach across adapters but I definitely want to tackle this

dbt/adapters/bigquery/connections.py
```diff
@@ -713,8 +714,24 @@ def _query_and_results(
                 self._bq_job_link(query_job.location, query_job.project, query_job.job_id)
             )
 
-        iterator = query_job.result(max_results=limit, timeout=job_execution_timeout)
+        # only use async logic if user specifies a timeout
+        if job_execution_timeout:
+            loop = asyncio.new_event_loop()
+            future_iterator = asyncio.wait_for(
+                loop.run_in_executor(None, query_job.result), timeout=job_execution_timeout
```
mikealfare, line 721: What happened to the `max_results=limit` kwarg? It's still in the non-`job_execution_timeout` route and I don't think the usage of it correlates with providing a timeout.

dbt/adapters/bigquery/connections.py
```diff
@@ -713,8 +714,24 @@ def _query_and_results(
                 self._bq_job_link(query_job.location, query_job.project, query_job.job_id)
             )
 
-        iterator = query_job.result(max_results=limit, timeout=job_execution_timeout)
+        # only use async logic if user specifies a timeout
+        if job_execution_timeout:
```
mikealfare, line 718: This could be overkill, but I could see an argument for putting all of this new code into a private method so that the code here becomes something like:
```python
if job_execution_timeout:
    iterator = self._run_timeout_query(query_job, timeout, limit)
else:
    iterator = query_job.result(max_results=limit)
return query_job, iterator
```

dbt/adapters/bigquery/connections.py
```diff
@@ -713,8 +714,24 @@ def _query_and_results(
...
+            loop = asyncio.new_event_loop()
+            future_iterator = asyncio.wait_for(
+                loop.run_in_executor(None, query_job.result), timeout=job_execution_timeout
+            )
 
+            try:
+                iterator = loop.run_until_complete(future_iterator)
+            except asyncio.TimeoutError:
+                query_job.cancel()
+                raise DbtDatabaseError(
```
mikealfare, line 728: Is this necessarily a `DbtDatabaseError`? There are several reasons this could time out, including user error (e.g. cartesian join) and server side error. It might be right, I'm asking for confirmation; I don't actually know what `DbtDatabaseError` covers.

tests/functional/test_job_timeout.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @pytest.fixture(scope="class")
+    def profiles_config_update(self, dbt_profile_target):
+        outputs = {"default": dbt_profile_target}
+        outputs["default"]["job_execution_timeout_seconds"] = _SHORT_TIMEOUT
+        return {"test": {"outputs": outputs, "target": "default"}}
+
+    def test_job_timeout(self, project):
+        result = run_dbt(["run"], expect_pass=False)  # project setup will fail
+        assert f"Query exceeded configured timeout of {_SHORT_TIMEOUT}s" in result[0].message
```
mikealfare, line 62: Could we also verify that it throws the type of error that we're expecting above (currently `DbtDatabaseError`)?

dbt/adapters/bigquery/connections.py
colin-rogers-dbt: yea I'm not sure either, this could also probably be it's own exception eventually `DbtModelTimeout` or something

dbt/adapters/bigquery/connections.py
colin-rogers-dbt: I went back and forth on that, what I'd really like to do is abstract all of this out of the class to separate function. Agreed that this feels like a it's own method but I'm not sure if a private method would really help or just add more clutter to an already overloaded class

mikealfare (approved):

tests/functional/test_job_timeout.py
colin-rogers-dbt: Somewhat duplicative as we validate the kind of exception in unit testing but can't hurt

Issue #917, opened 2023-09-08
[ADAP-879] [Feature] Query Cancellation


d-cole:
### Is this your first time submitting a feature request?

- [X] I have read the [expectations for open source contributors](https://docs.getdbt.com/docs/contributing/oss-expectations)
- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing dbt-bigquery functionality, rather than a Big Idea better suited to a discussion

### Describe the feature

When a dbt-bigquery run is interrupted running queries should be canceled.

### Describe alternatives you've considered

Relying on timeouts possibly added by this [PR](https://github.com/dbt-labs/dbt-bigquery/pull/902/). While timeouts can help, it does not solve the issue of wasted resources for local query cancelation.

### Who will this benefit?

 This will benefit users of dbt-bigquery in two scenarios:
 - When executions triggered during local development are interrupted, this feature will avoid wasting compute. This will help reduce cost, especially when users are using on-demand billing. Additionally, it will save users time as they will not need to use the BQ ui and api to identify and cancel query jobs.

 - This will allow for orchestrator-configured timeouts (e.g. airflow) to safely cancel model execution. This will save compute and ensure retries don't rerun a query that was unknowingly completed or still running.

### Are you interested in contributing this feature?

Yes https://github.com/dbt-labs/dbt-bigquery/pull/918

github-christophe-oudar:
Great feature 👍 
One thing I'd like to see added on top of the `job_id` set up is to be able to override that id generation with a Jinja macro.
Creating the id (with the UUID) and then passing to the Jinja to transform it, would enable to prefix/suffix the `job_id` with `dbt_` for instance (to easily spot dbt jobs) or even further add the model name!
WDYT?

colin-rogers-dbt:
Ideally this could be handled in dbt-core, have opened this discussion: https://github.com/dbt-labs/dbt-core/discussions/8522

d-cole:
> Ideally this could be handled in dbt-core, have opened this discussion: [dbt-labs/dbt-core#8522](https://github.com/dbt-labs/dbt-core/discussions/8522)

@colin-rogers-dbt 
[...]

PR #918, opened 2023-09-08
Add Query Cancellation

d-cole:
resolves # https://github.com/dbt-labs/dbt-bigquery/issues/917
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/# N/A

### Problem

The BigQueryConnectionManager cannot cancel queries.

### Solution

This PR adds query cancellation by:
 - Predetermining and storing job_ids on a per-thread basis in BigQueryConnectionManager.job_ids_by_thread
 - Implements cancel_open to attempt to cancel any running (or previously run) job ids associated with each thread before closing the connection. This is done by using the existing `_retry_and_handle` to call [client.cancel_job](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/client.py#L2108).

**Why predetermine job_ids?**
Initially I considered persisting job_ids after [client.query is called](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/adapters/bigquery/connections.py#L712). However, this could lead to a race condition if a job has been started and a termination is sent before the job_id was stored, leading to a failure to cancel the job. By predetermining job_ids (uuid4), we can persist the job_id before the job has been kicked off. Doing this, the race condition only leads to attempting to cancel a job that doesn't exist.

**Alternatives considered:**
I considered using sessions similar to the [Snowflake adapter](https://github.com/dbt-labs/dbt-snowflake/blob/e72e8f1fd569a40cd3ffabcdcd3968f3bd0d8acc/dbt/adapters/snowflake/connections.py#L381). This doesn't seem feasible as the python api currently does not support [session termination](https://cloud.google.com/bigquery/docs/sessions-terminating).

**Questions for reviewers:**
 - Are there concerns around modifying jobs_by_thread without locking? 
 - What (if any) are the implications of this for python models?
 - When a job has completed it is not removed from jobs_by_thread. This means upon cancelation it will attempt to cancel jobs that have already completed. Will this cause any issues?
 - What logging should be added upon the successful cancelation of a job?
 - What integration tests (if any) are needed?

### Checklist


jtcohen6:
Thanks @d-cole - we can try taking another look here. The changes in this PR are fairly intricate, so it will require careful review — but it's also true that when dbt is cancelling queries it's because something has already gone wrong (timeout, `--fail-fast`, etc).

colin-rogers-dbt:
closing in favor of #1251

PR #840, opened 2023-07-21
738/wazi

wazi55:
resolves #
[docs](https://github.com/dbt-labs/docs.getdbt.com/issues/new/choose) dbt-labs/docs.getdbt.com/#

https://github.com/dbt-labs/dbt-bigquery/issues/734

### Problem

Serverless Spark polling within the dbt-bigquery adapter spends an extra 1.5 minute waiting for cluster to be torn down instead of just waiting for the job to finish. 

### Solution

Updated operation.result to getBatch instead. added a randomly generated batch_id to allow dataproc to reference getBatch operation from the batch_id used for createBatch. 

### Checklist

- [ x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #929, opened 2023-09-19
poll `.GetBatch()` instead of using `operation.result()`

colin-rogers-dbt:
- re:PR https://github.com/dbt-labs/dbt-bigquery/pull/840
- adding back comment # check if job failed
- adding changelog
- precommit code format
- sleep(2) first in the while loop before the request to eliminate the last 2 seconds sleep if the response is in one of the 3 options
- removing empty spaces
- update batch request to handle `GetBatchRequest`

resolves #

### Checklist


tests/functional/adapter/test_python_model.py
```diff
@@ -66,7 +66,7 @@ def model(dbt, spark):
 """
 
 models__python_array_batch_id_python = """
-import pandas
+import pandas as pd
```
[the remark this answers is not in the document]
colin-rogers-dbt: this might be due to a core change since we don't run these in CI unfortunately but can confirm the test fails if we don't alias as od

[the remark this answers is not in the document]
mikealfare: Ah ok, so it was already failing, but because we don't run them in CI, we didn't know that. Makes sense now.

dbt/adapters/bigquery/python_submissions.py
```diff
@@ -118,20 +121,33 @@ def _get_job_client(self) -> dataproc_v1.BatchControllerClient:
     def _get_batch_id(self) -> str:
         return self.parsed_model["config"].get("batch_id")
 
-    def _submit_dataproc_job(self) -> dataproc_v1.types.jobs.Job:
+    def _submit_dataproc_job(self) -> Batch:
```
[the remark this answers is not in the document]
colin-rogers-dbt: I think adding a new State enum introduces complexity we don't really need but I agree on breaking this up a bit.

dbt/adapters/bigquery/dataproc/batch.py
```diff
@@ -0,0 +1,63 @@
...
+        state = response.state
+    if not response:
+        raise ValueError("No response from Dataproc")
+    if run_time >= timeout:
+        raise ValueError(
+            f"Operation did not complete within the designated timeout of {timeout} seconds."
+        )
+    if state != Batch.State.SUCCEEDED:
+        raise ValueError(response.state_message)
+    breakpoint()
```
mikealfare, line 47: Whoops

dbt/adapters/bigquery/dataproc/batch.py
```diff
@@ -0,0 +1,63 @@
...
+    state = Batch.State.PENDING
+    response = None
+    run_time = 0
+    while state in _BATCH_RUNNING_STATES and run_time < timeout:
+        time.sleep(1)
+        response = job_client.get_batch(  # type: ignore
+            request=GetBatchRequest(name=batch_name),  # type: ignore
+        )
+        run_time = datetime.now().timestamp() - response.create_time.timestamp()  # type: ignore
+        state = response.state
```
mikealfare, line 38: Do we need to check if `response` is `None` here? We're throwing an error for that condition after we exit the while loop, so it seems like a possibility. This would throw an attribute error in that case, prior to getting to the error check.

Same issue with `response.create_time.timestamp()` above.

dbt/adapters/bigquery/dataproc/batch.py
```diff
@@ -0,0 +1,63 @@
...
+    if not response:
+        raise ValueError("No response from Dataproc")
+    if run_time >= timeout:
+        raise ValueError(
+            f"Operation did not complete within the designated timeout of {timeout} seconds."
+        )
+    if state != Batch.State.SUCCEEDED:
```
mikealfare, line 45: Should we check happy path first and return if we're there? Is there a scenario (other than finishing successfully .5s after `timeout`) where `state == Batch.State.SUCCEEDED` but we want to raise an error instead of returning `response`?

dbt/adapters/bigquery/dataproc/batch.py
```diff
@@ -0,0 +1,63 @@
...
+)
+from google.protobuf.json_format import ParseDict
+
+from dbt.adapters.bigquery.connections import DataprocBatchConfig
+
+_BATCH_RUNNING_STATES = [Batch.State.PENDING, Batch.State.RUNNING]
+DEFAULT_JAR_FILE_URI = "gs://spark-lib/bigquery/spark-bigquery-with-dependencies_2.12-0.21.1.jar"
+
+
+def create_batch_request(batch: Batch, batch_id, project, region) -> CreateBatchRequest:
```
mikealfare, line 19: nit: missing type annotations in signature

dbt/adapters/bigquery/dataproc/batch.py
```diff
@@ -0,0 +1,63 @@
...
+    return CreateBatchRequest(
+        parent=f"projects/{project}/locations/{region}",  # type: ignore
+        batch_id=batch_id,
+        batch=batch,  # type: ignore
+    )
+
+
+def poll_batch_job(parent, batch_id, job_client: BatchControllerClient, timeout: int) -> Batch:
```
mikealfare, line 27: nit: missing type annotations in signature

dbt/adapters/bigquery/dataproc/batch.py
colin-rogers-dbt: No we don't need to check here since the call will return a response or raise an exception

colin-rogers-dbt: tbh the `if not response` check isn't _really_ needed but it makes the type checker happy

dbt/adapters/bigquery/dataproc/batch.py
colin-rogers-dbt: moved the timeout check to be nested under the succeeded check

mikealfare (approved):

PR #954, opened 2023-10-05
Use branch 8007-contract_type_aliasing

### LINKED ISSUE #953: Support dbt-core 8592 type aliasing in model contracts
### Description

With pull request https://github.com/dbt-labs/dbt-core/pull/8592 the behavior of types in contracts changed. We need to tweak a test case to match the change in behavior.

### Solution

Update the test case.

gshank:
resolves #953

### Problem

Need to match changes in https://github.com/dbt-labs/dbt-core/pull/8592

### Solution

Fix the test.   Add TEXT => STRING conversion to TYPE_LABELS.

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/column.py
+++ b/dbt/adapters/bigquery/column.py
@@ -13,12 +13,9 @@ Self = TypeVar("Self", bound="BigQueryColumn")
 @dataclass(init=False)
 class BigQueryColumn(Column):
     TYPE_LABELS = {
-        "STRING": "STRING",
-        "TIMESTAMP": "TIMESTAMP",
+        "TEXT": "STRING",
         "FLOAT": "FLOAT64",
         "INTEGER": "INT64",
-        "BOOLEAN": "BOOLEAN",
-        "RECORD": "RECORD",
     }
     fields: List[Self]  # type: ignore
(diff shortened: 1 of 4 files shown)


dbt/adapters/bigquery/column.py
```diff
@@ -13,12 +13,9 @@
 @dataclass(init=False)
 class BigQueryColumn(Column):
     TYPE_LABELS = {
-        "STRING": "STRING",
```
colin-rogers-dbt, line 16: Shouldn't we just be adding `TEXT`? Why delete the other types?

gshank: There is no point to types that are the same on both sides. They do nothing except (in some cases) change the type from lowercase to uppercase.

gshank: I removed them in the dbt-core PR because they were causing unnecessary churning in the test cases.

gshank:
The test is currently failing because in test_constraints.py the "integer" type is being translated to "int64". It looks like the tests were working before with "integer". Is that conversion not necessary?

colin-rogers-dbt (approved):

PR #955, opened 2023-10-05
[fix] Use rendered query comment for job labels

kodaho:
resolves #863

### Problem

A regression appears in the last release of `dbt-bigquery`. A user can previously add labels to a BigQuery job thanks to a macro returning a mapping. In the version 1.6.0, this was no longer possible even if the project is set to activate this feature. 

### Solution

The proposed solution is to use, instead of `self.profile.query_comment` (which is the not-rendered Jinja macro), `self.query_header.comment.query_comment` (the JSON rendered query comments).

I added some safeguards to make sure that `self.query_header.comment.query_comment` is not null but during the runtime, it should never be the case because of the following succession of functions and methods calls during a run command:
1. The query labels are sent with a `run` command, [command decorated with the decorator `requires.manifest`](https://github.com/dbt-labs/dbt-core/blob/f30293359ce709d7b017b09482124b72e8013b0e/core/dbt/cli/main.py#L589).
2. The class method [`ManifestLoader.get_full_manifest`](https://github.com/dbt-labs/dbt-core/blob/f30293359ce709d7b017b09482124b72e8013b0e/core/dbt/cli/requires.py#L271) is called.
3. The method [`ManifestLoader().save_macros_to_adapter`](https://github.com/dbt-labs/dbt-core/blob/f30293359ce709d7b017b09482124b72e8013b0e/core/dbt/parser/manifest.py#L324C20-L324C42) is called.
4. `ManifestLoader().macro_hook` is called, knowing that [`macro_hook = adapter.connections.set_query_header`](https://github.com/dbt-labs/dbt-core/blob/f30293359ce709d7b017b09482124b72e8013b0e/core/dbt/parser/manifest.py#L287).
5. Calling `Connection().set_query_header` makes sure that `self.query_header` is not null.

### Checklist

- [x] I have read [the contributing guide](https://github.com/dbt-labs/dbt-core/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [x] I have run this code in development and it appears to resolve the stated issue
- [x] This PR includes tests, or tests are not required/relevant for this PR
- [x] This PR has no interface changes (e.g. macros, cli, logs, json artifacts, config files, adapter interface, etc) or this PR has already received feedback and approval from Product or DX

PR #957, opened 2023-10-06
a slight rework of 955 to keep a seperation of concerns

McKnight-42:
resolves #

### Problem
we implemented a change in #955 to fix query_header in doing so moved `self.query_header` into the same conditional logic as `self.profile` and used a _walrus_ operator to assign and check which while clean does make parse code quickly slightly harder.

### Solution
to functionalize the small snippet and create a small change by making it to two nested if statements and moving the empty set of variable to top of function.

### Checklist


VersusFacit (approved):
Big fan of this assuming the tests are all happy and you've tested the semantics :) The function gets us bonus documentation as to what/why is happening!

Thanks for being detail oriented!

PR #977, opened 2023-10-23
Use deterministic job_ids to avoid retrying successful queries

### LINKED ISSUE #949: [ADAP-924] [Feature] Use deterministic job_ids to avoid retrying successful queries

### Describe the feature

Currently if we experience a transient exception like `RemoteDisconnected` we can sometimes end up re-running a query that has been successfully kicked off. 

We could use a deterministic job_id based off of the run id and model name to make retry calls idempotent. If a query already exists for that job_id we can poll for it's status instead of rerunning the query. 

_No response_

### Who will this benefit?

_No response_

_No response_

### Anything else?

_No response_

McKnight-42:
resolves #949 

### Problem
Currently if we experience a transient exception like RemoteDisconnected we can sometimes end up re-running a query that has been successfully kicked off.

### Solution
on retry try to poll by a deterministic job_id to see if a successful job has already been kicked off

### Checklist


dbt/adapters/bigquery/connections.py
```diff
@@ -283,7 +288,26 @@ def exception_handler(self, sql):
...
+
+                        def fn():
+                            return client.cancel_job(job_id)
+
+                        self._retry_and_handle(msg=f"Cancel job: {job_id}", conn=connection, fn=fn)
+
+                    self.close(connection)
+
+                if connection.name is not None:
+                    names.append(connection.name)
```
McKnight-42, line 310: Do we need to return `names` here? if we do we need to re-add it and change the type for function.

tests/unit/test_bigquery_adapter.py
```diff
@@ -340,12 +345,12 @@ def test_acquire_connection_maximum_bytes_billed(self, mock_open_connection):
...
         adapter = self.get_adapter("oauth")
-        self.assertEqual(adapter.cancel_open_connections(), None)
+        self.assertEqual(len(list(adapter.cancel_open_connections())), 0)
 
     def test_cancel_open_connections_master(self):
         adapter = self.get_adapter("oauth")
-        adapter.connections.thread_connections[0] = object()
-        self.assertEqual(adapter.cancel_open_connections(), None)
+        adapter.connections.thread_connections[0] = mock_connection("master")
+        self.assertEqual(len(list(adapter.cancel_open_connections())), 0)
```
McKnight-42, line 353: this is currently failing as the len is showing 1 meaning a job isn't being dropped because its an empty list because this is testing dropping connections not jobs is there a loophole we need to add to our above logic in `cancel_open`

dbt/adapters/bigquery/connections.py
```diff
@@ -239,6 +240,10 @@ class BigQueryConnectionManager(BaseConnectionManager):
     DEFAULT_INITIAL_DELAY = 1.0  # Seconds
     DEFAULT_MAXIMUM_DELAY = 3.0  # Seconds
 
+    def __init__(self, profile: AdapterRequiredConfig):
+        super().__init__(profile)
+        self.jobs_by_thread: Dict[Any, Any] = {}
```
McKnight-42, line 245: do we have any more stricter sense we want here should this just be str?

colin-rogers-dbt: I'm pretty sure this can just be `str, str`

tests/unit/test_bigquery_adapter.py
colin-rogers-dbt: I think it's the opposite, since this is now implemented we _are_ dropping connections so the length is now 1 (it was 0 since no connections were being dropped before). 
This is pretty confusing since the base adapter assumes a ODBC/JDBC style of connecting to a database (i.e. where you hold a connection open to a database) so a thread is equivalent to a connection. Instead with bigquery we make requests and get async jobs.

McKnight-42: okay got the three unit tests passing, had to incorporate some stuff from other repos and go over how we build out the `add_query` part for bigquery since it doesn't actually have that method in `connections`

My understanding of the original comment is the above was failing as we weren't targeting `key` thread identifier variable on the mock so wasn't correctly dropping in the assert

McKnight-42:
all integration/functional tests failing now for 
```
dbt.exceptions.DbtRuntimeError: Runtime Error
E             'Client' object has no attribute 'connection'
```

McKnight-42:
> all integration/functional tests failing now for
> 
> ```
> dbt.exceptions.DbtRuntimeError: Runtime Error
> E             'Client' object has no attribute 'connection'
> ```
> 
> in `raw_execute`

solved this part of it by not going through defined client and just using conn.name

McKnight-42:
@colin-rogers-dbt it's possible we may still have to use `uuid` like the original pr as I don't think `invocation_id` is active on a fresh run during the `raw_execute` so we may need to define a unique identifier instead. thoughts?

dbt/adapters/bigquery/connections.py
```diff
@@ -292,8 +298,29 @@ def exception_handler(self, sql):
...
+                    for job_id in self.jobs_by_thread.get(thread_id, []):
+
+                        def fn():
+                            return client.cancel_job(job_id)
+
+                        self._retry_and_handle(msg=f"Cancel job: {job_id}", conn=connection, fn=fn)
+
+                    self.close(connection)
+
+                if connection.name is not None:
```
mikealfare, line 320: Do we specifically want all connections which are not `this_connection`? Or do we only want connections in which we cancelled jobs? In this current flow, a connection for which `connection.state == ConnectionState.CLOSED` will show up in `names`, which doesn't feel like an intuitive list to get from `cancel_open'.

dbt/adapters/bigquery/connections.py
```diff
@@ -292,8 +298,29 @@ def exception_handler(self, sql):
...
-        pass
+    def cancel_open(self) -> List[str]:
+        names = []
+        this_connection = self.get_if_exists()
+        with self.lock:
+            for thread_id, connection in self.thread_connections.items():
+                if connection is this_connection:
+                    continue
+
+                if connection.handle is not None and connection.state == ConnectionState.OPEN:
```
mikealfare, line 309: This could be over-engineering, but I would consider putting the contents of this if block into its own method.

dbt/adapters/bigquery/connections.py
```diff
@@ -491,18 +518,26 @@ def raw_execute(
 
         job_creation_timeout = self.get_job_creation_timeout_seconds(conn)
         job_execution_timeout = self.get_job_execution_timeout_seconds(conn)
+        # build out determinsitic_id
+        model_name = conn.credentials.schema  # schema name as model name is not
+        invocation_id = str(uuid.uuid4())
+        job_id = define_job_id(model_name, invocation_id)
```
mikealfare, line 524: I don't think `uuid.uuid4()` is deterministic, which means `job_id` is not either. Have you considered an `md5` hash of sufficient attributes (model, connection name, etc.)?

dbt/adapters/bigquery/jobs.py
```diff
@@ -0,0 +1,3 @@
+def define_job_id(model_name, invocation_id):
+    job_id = f"{model_name}_{invocation_id}"
```
mikealfare, line 2: What's the constraint on `job_id`? Is there a max length? Can all characters that go into a model name also be used in a job id?

dbt/adapters/bigquery/connections.py
McKnight-42: currently calling `uuid` directly as part of getting unit tests swapped over for functionality I think initial/current plan was to use the `invocation_id` we define via tracking in core https://docs.getdbt.com/reference/dbt-jinja-functions/invocation_id and it itself is a `uuid` based on docs.

dbt/adapters/bigquery/connections.py
McKnight-42: If i backtrack `open_cancel` to core it looks like one of the only places we call it is for `cancel_open_connections` which does make me think the desired result is that all closed connections are accounted for?

dbt/adapters/bigquery/connections.py
McKnight-42: to be clear are you referring to lines 308-321?

dbt/adapters/bigquery/jobs.py
McKnight-42: will definitely have to test this, I think all characters are fine as we should just be combining 2 strings but length may hit a limit

dbt/adapters/bigquery/connections.py
mikealfare: I'm referring to lines 310-318. Then this would look like:

```python
names = []
this_connection = self.get_if_exists()

with self.lock:
    for thread_id, connection in self.thread_connections.items():

        if connection is this_connection:
            continue

        if connection.handle and connection.state == ConnectionState.OPEN:
            self.close_thread(thread_id, connection)  # or whatever name you choose

        if name := connection.name:
            names.append(name)

return names
```

Something else worth considering is whether we want to handle all of the threads within a connection. I don't know if there's more than one thread for a connection, but I feel like there is. If there's a connection with more than one thread, you'll close that connection in the second condition above when you get to the first thread. Then you'll skip past the second condition for every other thread since `connection.state` should be closed at that point.

I think what you probably want is a list of job_ids by connection. Then for each connection you would cancel the job. Once all jobs are cancelled, then close the connection.

dbt/adapters/bigquery/connections.py
mikealfare: In other words we should only be returning the connections which we cancelled during this call, right?

dbt/adapters/bigquery/jobs.py
mikealfare: I would want to make sure that we can submit a job_id to BQ with some weird characters. People put all kinds of things in their model names. An alternative is to hash the model_name so that it's only alpha-numeric.

colin-rogers-dbt: +1 if we want to stick with uuid we can just generate a deterministic one with `uuid.uuid5`

dbt/adapters/bigquery/connections.py
colin-rogers-dbt: Instead of using invocation_id (which we only sometimes have) should we just use the actual query text (which we have to have)?

dbt/adapters/bigquery/jobs.py
```diff
@@ -0,0 +1,12 @@
...
+_INVOCATION_ID = uuid.uuid4()
+
+
+def define_job_id(sql, invocation_id=None):
+    if invocation_id:
+        job_id = str(uuid.uuid5(invocation_id, sql))
+    else:
+        job_id = str(uuid.uuid5(_INVOCATION_ID, sql))
+    job_id = job_id.replace("-", "_")
+    return job_id
```
github-christophe-oudar, line 12: I would leverage a macro to let end users override that logic to make it unique across invocations if needed

PR #986, opened 2023-10-27
Fix merge fails when partition is required

### LINKED ISSUE #792: [ADAP-659] [CT-2745] [Bug] BigQuery merge fails when partition is required

### Current Behavior

Trying to migrate a BQ stored procedure to DBT where the partition on the target table is required. I receive this error on the table the requires a partition:

> BadRequest("
>     Cannot query over table 'test' without a filter over column(s) 'date' that can be used for partition elimination")

### Expected Behavior

I would expect the ability to manage a table in such a way similar to how we mange `insert_overwrite` partitioning where I can provide a list of dates say, last 14 days, and have the target filtered and merge on that look back period. 
I cannot use `insert_overwrite` at the moment because I am not merging from the whole set of data, so I would lose data that has not been updated when the merge should take care of not losing an unmatched ID. 

### Steps To Reproduce

Build this twice to reproduce, second will be failure. I have tried on `1.4` and `1.5` to no avail.

```sql
{{
    config(
        materialized="incremental",
        incremental_strategy="merge",
        partition_by={
            "field": "date",
            "data_type": "DATE",
            "granularity": "day",
        },
        unique_key="value",
        require_partition_filter=true,
    )
}}

select
    current_date() as date,
     1 as value
```

### Relevant log output

_No response_

### Environment

```markdown
- dbt-cloud: 1.4
```

bigquery

### Additional Context

_No response_

tnk-ysk:
resolves #792

### Problem
Trying to migrate a BQ stored procedure to DBT where the partition on the target table is required. I receive this error on the table the requires a partition:

BadRequest("
Cannot query over table 'test' without a filter over column(s) 'date' that can be used for partition elimination")

### Solution
Avoid require partition filter by adding partition is null or partition is not null to the merge condition.
This condition does not change the target record.

### Checklist


tnk-ysk:
I'm wondering if I should add the partition_filld parameter to [get_merge_sql](https://github.com/dbt-labs/dbt-core/blob/a2a7b7d795eb40f743063601375e44a5b72f1b9a/core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql#L1) and modify [default__get_merge_sql](https://github.com/dbt-labs/dbt-core/blob/a2a7b7d795eb40f743063601375e44a5b72f1b9a/core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql#L7) instead of adding bigquery__get_merge_sql.
Please let us know what you think.

github-christophe-oudar:
Great contribution @tnk-ysk!
Could it be possible to extract 
```
{% if partition_config and config.get('require_partition_filter') -%}
            {% set avoid_require_partition %}
                DBT_INTERNAL_DEST.{{ partition_config.field }} IS NULL OR DBT_INTERNAL_DEST.{{ partition_config.field }} IS NOT NULL
            {% endset %}
            {% do predicates.append(avoid_require_partition) %}
        {%- endif -%}
```

And add that condition to the predicates so that you can combine it with https://github.com/dbt-labs/dbt-bigquery/pull/994/files?

tnk-ysk:
@github-christophe-oudar 
Thank you for your review.
I extracted the process and made it into a macro, so please check it.

I will to fix #994 after this PR is merged.

github-christophe-oudar:
@tnk-ysk it looks better! 
Then wouldn't you be able to keep the default macro and pass the `predicates` array as the `incremental_predicates` param from the `get_merge_sql` macro?
My point is that if someone has overridden the `get_merge_sql` macro inside its dbt project for any reason, he won't benefit from the change if we redefine it that way.

tnk-ysk:
@github-christophe-oudar
Thank you for your wonderful comment.
I fixed it and also made the macro return the predicates instead of making destructive change.

github-christophe-oudar (approved):
LGTM, I'll let dbt Labs team takes over the review as I don't have any rights.

.changes/unreleased/Fixes-20231105-143145.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Fixes
+body: Merge fails when partition is required
```
colin-rogers-dbt, line 2:
```suggestion
body: In incremental models, add dummy merge condition on source partition column when partition is required
```

colin-rogers-dbt (approved):

PR #996, opened 2023-10-31
ADAP-972: Fix issue where materialized views were being mapped as tables in catalog queries

### LINKED ISSUE #995: [ADAP-972] [Bug] Materialized views show up as tables in catalog queries

### Current Behavior

Materialized views show up as tables in catalog queries.

### Expected Behavior

Materialized views should show up as materialized views in catalog queries.

### Steps To Reproduce

1. Create a materialized view
2. Run `docs generate`
3. Notice that the type for the materialized view is "table"

For more details, refer to: https://github.com/dbt-labs/dbt-core/issues/8864

### Relevant log output

_No response_

### Environment

```markdown
- OS:
- Python:
- dbt-core:
- dbt-bigquery:
```

### Additional Context

_No response_

mikealfare:
resolves #995, dbt-labs/dbt-core#8864

### Problem

Materialized views were being mapped as tables in catalog queries.

### Solution

- write a test demonstrating the issue
- update the catalog query to return the correct type

### Checklist

### DIFF
--- a/dbt/include/bigquery/macros/catalog.sql
+++ b/dbt/include/bigquery/macros/catalog.sql
@@ -7,30 +7,42 @@
   {%- else -%}
 
   {%- set query -%}
-    with tables as (
+    with materialized_views as (
         select
-            project_id as table_database,
-            dataset_id as table_schema,
-            table_id as original_table_name,
+            table_catalog as project_id,
+            table_schema as dataset_id,
+            table_name as table_id
+        from {{ information_schema.replace(information_schema_view='MATERIALIZED_VIEWS') }}
+    ),
+    tables as (
+        select
+            tables.project_id as table_database,
+            tables.dataset_id as table_schema,
+            tables.table_id as original_table_name,
 
-            concat(project_id, '.', dataset_id, '.', table_id) as relation_id,
+            concat(tables.project_id, '.', tables.dataset_id, '.', tables.table_id) as relation_id,
 
-            row_count,
-            size_bytes as size_bytes,
+            tables.row_count,
+            tables.size_bytes as size_bytes,
             case
-                when type = 1 then 'table'
-                when type = 2 then 'view'
+                when materialized_views.table_id is not null then 'materialized view'
+                when tables.type = 1 then 'table'
+                when tables.type = 2 then 'view'
                 else 'external'
             end as table_type,
 
-            REGEXP_CONTAINS(table_id, '^.+[0-9]{8}$') and coalesce(type, 0) = 1 as is_date_shard,
-            REGEXP_EXTRACT(table_id, '^(.+)[0-9]{8}$') as shard_base_name,
-            REGEXP_EXTRACT(table_id, '^.+([0-9]{8})$') as shard_name
+            REGEXP_CONTAINS(tables.table_id, '^.+[0-9]{8}$') and coalesce(type, 0) = 1 as is_date_shard,
+            REGEXP_EXTRACT(tables.table_id, '^(.+)[0-9]{8}$') as shard_base_name,
+            REGEXP_EXTRACT(tables.table_id, '^.+([0-9]{8})$') as shard_name
 
-        from {{ information_schema.replace(information_schema_view='__TABLES__') }}
+        from {{ information_schema.replace(information_schema_view='__TABLES__') }} tables
+        left join materialized_views
+            on materialized_views.project_id = tables.project_id
+            and materialized_views.dataset_id = tables.dataset_id
+            and materialized_views.table_id = tables.table_id
         where (
           {%- for schema in schemas -%}
-            upper(dataset_id) = upper('{{ schema }}'){%- if not loop.last %} or {% endif -%}
+            upper(tables.dataset_id) = upper('{{ schema }}'){%- if not loop.last %} or {% endif -%}
           {%- endfor -%}
         )
     ),
--- a/tests/functional/adapter/catalog_tests/test_relation_types.py
+++ b/tests/functional/adapter/catalog_tests/test_relation_types.py
@@ -37,7 +37,7 @@ class TestCatalogRelationTypes:
         self, docs: CatalogArtifact, node_name: str, relation_type: str
     ):
         """
-        This test addresses: https://github.com/dbt-labs/dbt-core/issues/8864
+        This test addresses: https://github.com/dbt-labs/dbt-bigquery/issues/995
         """
         assert node_name in docs.nodes
         node = docs.nodes[node_name]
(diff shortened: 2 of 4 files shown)


dbt/include/bigquery/macros/catalog.sql
```diff
@@ -7,30 +7,42 @@
...
+                when materialized_views.table_id is not null then 'materialized view'
+                when tables.type = 1 then 'table'
+                when tables.type = 2 then 'view'
                 else 'external'
             end as table_type,
 
-            REGEXP_CONTAINS(table_id, '^.+[0-9]{8}$') and coalesce(type, 0) = 1 as is_date_shard,
-            REGEXP_EXTRACT(table_id, '^(.+)[0-9]{8}$') as shard_base_name,
-            REGEXP_EXTRACT(table_id, '^.+([0-9]{8})$') as shard_name
+            REGEXP_CONTAINS(tables.table_id, '^.+[0-9]{8}$') and coalesce(type, 0) = 1 as is_date_shard,
```
VersusFacit, line 34: These regexes make me uncomfortable. Is there any way to pull them out as vars or something so we can at least document them a la C macros? If nothing else, it might be nice to know those that should be the same, do in fact stay the same

tests/functional/adapter/catalog_tests/test_relation_types.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class TestCatalogRelationTypes:
+    @pytest.fixture(scope="class", autouse=True)
+    def seeds(self):
+        return {"my_seed.csv": files.MY_SEED}
+
+    @pytest.fixture(scope="class", autouse=True)
+    def models(self):
+        yield {
```
VersusFacit, line 15: What's up with `yield` over return.

Also, really like the flow this test. Points for aesthetic.

dbt/include/bigquery/macros/catalog.sql
```diff
@@ -7,30 +7,42 @@
...
+            concat(tables.project_id, '.', tables.dataset_id, '.', tables.table_id) as relation_id,
 
-            row_count,
-            size_bytes as size_bytes,
+            tables.row_count,
+            tables.size_bytes as size_bytes,
             case
-                when type = 1 then 'table'
-                when type = 2 then 'view'
+                when materialized_views.table_id is not null then 'materialized view'
```
VersusFacit, line 28: given that the config block, should this be `materialized_view` instead of here with the space. A nit only. Just thinking about consistency of output.

dbt/include/bigquery/macros/catalog.sql
```diff
@@ -7,30 +7,42 @@
...
-            REGEXP_CONTAINS(table_id, '^.+[0-9]{8}$') and coalesce(type, 0) = 1 as is_date_shard,
-            REGEXP_EXTRACT(table_id, '^(.+)[0-9]{8}$') as shard_base_name,
-            REGEXP_EXTRACT(table_id, '^.+([0-9]{8})$') as shard_name
+            REGEXP_CONTAINS(tables.table_id, '^.+[0-9]{8}$') and coalesce(type, 0) = 1 as is_date_shard,
+            REGEXP_EXTRACT(tables.table_id, '^(.+)[0-9]{8}$') as shard_base_name,
+            REGEXP_EXTRACT(tables.table_id, '^.+([0-9]{8})$') as shard_name
 
-        from {{ information_schema.replace(information_schema_view='__TABLES__') }}
+        from {{ information_schema.replace(information_schema_view='__TABLES__') }} tables
+        left join materialized_views
```
VersusFacit, line 39: I honestly don't know the answer here: Why is `left join` the right answer? Is your thinking that some tables will not have a materialized view entry to match it (i.e. they aren't materialized views even though we see them as tables).

VersusFacit (commented):
Several nits around Jinja implementation. 

Logic seems good since you're just adding on an extra information schema call and differentiating that set from the existing set. 

Great test with exhaustive but clear use cases. That `yield` for the docs fixture is magical

dbt/include/bigquery/macros/catalog.sql
mikealfare: They were already there. Those lines show up because I suffixed the column name with `tables.`. Perhaps we should revisit them (I share your concern), but that would be outside of the scope of this PR.

tests/functional/adapter/catalog_tests/test_relation_types.py
mikealfare: I suppose it doesn't matter here, as nothing happens after the `yield` in the fixture. The convention for a pytest fixture is:

```python
@pytest.fixture()
def my_cool_fixture(self):
    # do some setup pre-test
    yield some_value_you_want
    # do some teardown post-test
```

When the fixture simply returns a value, this can be simplified to:

```python
@pytest.fixture()
def my_cool_fixture(self):
    return some_value_you_want
```

I use `yield` out of habit of writing a pytest fixture, but both are valid.

dbt/include/bigquery/macros/catalog.sql
mikealfare: I thought about this in general for the three related PRs. It looks like we always return it with a space, but we return all caps for some adapters, all lowercase for others, and sometimes we return `"BASE TABLE"` whereas others return `"table"`. I feel like all of these should be standardized (probably to the `RelationType` enum), but I don't know if Cloud is depending on this list of values somehow. Given all of that, `'materialized view'` seemed the most consistent for now.

dbt/include/bigquery/macros/catalog.sql
mikealfare: That's correct. I'm left joining just to see if the record I'm looking at belongs to a materialized view. We need the types for all relations, not just materialized views; if we inner join, we'll drop all of the tables and views.

dbt/include/bigquery/macros/catalog.sql
VersusFacit: I can get behind that. They are clearing working right now. I just find the whole thing uncomfy as coding practice. However, since it's just a simple/intentional copy-paste job, I can get behind that.

tests/functional/adapter/catalog_tests/test_relation_types.py
VersusFacit: Wouldn't have guessed they were functionally equivalent. Weird then that we are using return in several places where yield does the job (and is more idiomatic). Oh well -- getting to know our tools! :)

dbt/include/bigquery/macros/catalog.sql
VersusFacit: you've clearly weighed the options which is what I was wondering most. I agree standardization is kind of a separate, high-level thing we'll need to address some day if the need presents itself. For now, I can live with the current state of affairs (even if from the outside I'd be asking 'wat')

VersusFacit (approved):

PR #1011, opened 2023-11-08
ADAP-464: Add allow_non_incremental_definition option to BigQuery materialized views

### LINKED ISSUE #672: Materialized Views - Non-incremental support
As [commented in the core MV thread](https://github.com/dbt-labs/dbt-core/issues/6911#issuecomment-1503925035), BigQuery is adding support for [non-incremental MVs](https://cloud.google.com/bigquery/docs/materialized-views-create#non-incremental).

From our side, it means that we have an additional setting to support for MVs in BQ: `allow_non_incremental_definition`: `true`/`false`. This setting is covered by `on_configuration_change` via DROP/CREATE since it can't be changed via ALTER.

When set, it requires `max_staleness` to be set. I'd rather leave BQ errors out here for now - this is a preview feature and I'm not sure this won't change before GA. We should not check this.

___
#### BigQuery

[BigQuery Materialized Views](https://cloud.google.com/bigquery/docs/materialized-views-intro)

- Configurations:
  - enable_refresh : BOOLEAN, optional (database level [default](https://cloud.google.com/bigquery/docs/materialized-views-manage#automatic_refresh) true) (see lifecycle for effect)
  - refresh_interval_minutes : integer, optional (database level [default](https://cloud.google.com/bigquery/docs/materialized-views-manage#automatic_refresh) at 30), only valid for enable_refresh true or default
  - max_staleness : time before bypassing the MV and running the statement instead (in preview but I like it!)
  - allow_non_incremental_definition : BOOLEAN, optional (database default `false`)
  - From models: [Partition / Cluster](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#using-table-partitioning-and-clustering), [TTL](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#controlling-table-expiration), [Labels](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#labels-and-tags)  
- Lifecycle : _default_ except Refresh is bypassed if auto_refresh
  - `on_configuration_change`.`apply` : enable_refresh/refresh_interval_minutes/max_staleness via [ALTER](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_materialized_view_set_options_statement), rest via DROP/[CREATE](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_materialized_view_statement)

bnaul:
Closes #672.

### Problem

Non-incremental materialized views are much more flexible than incremental and have a lot of uses: see e.g. https://medium.com/google-cloud/hidden-gems-of-bigquery-part-4-five-types-of-views-ddfbc05118fb#bc99

### Solution

Add `allow_non_incremental_definition` flag following the pattern for other fields. The only wrinkle is that the Python client library doesn't yet have a `property` for this variable, so it's being pulled directly from the `properties` dict instead.

I'm assuming the existing tests for materialized views will exercise this property as well (with the default False), if it seems necessary to add a special test with the flag set to True let me know.

### Checklist


dbt/adapters/bigquery/relation_configs/_options.py
```diff
@@ -21,6 +21,7 @@ class BigQueryOptionsConfig(BigQueryBaseRelationConfig):
     refresh_interval_minutes: Optional[float] = 30
     expiration_timestamp: Optional[datetime] = None
     max_staleness: Optional[str] = None
+    allow_non_incremental_definition: Optional[bool] = None
```
mikealfare, line 24: The database defaults to `False`, so we should do the same here so that we don't recognize it as a change each time.

dbt/adapters/bigquery/relation_configs/_options.py
```diff
@@ -137,6 +141,7 @@ def parse_bq_table(cls, table: BigQueryTable) -> Dict[str, Any]:
             "refresh_interval_minutes": table.mview_refresh_interval.seconds / 60,
             "expiration_timestamp": table.expires,
             "max_staleness": None,
+            "allow_non_incremental_definition": table._properties.get("materializedView", {}).get("allowNonIncrementalDefinition"),
```
mikealfare, line 144: We try to not use private attributes/methods (`_properties`) where we can. Is there another way to parse this from the `BigQueryTable` object?

bnaul: Unfortunately I don't think so, I mentioned this in the PR body

> The only wrinkle is that the Python client library doesn't yet have a property for this variable, so it's being pulled directly from the properties dict instead.

I'll make an issue on the `google-cloud-bigquery` repo as well, would you consider this a blocker or is it good enough to follow up and replace once it's added?

bnaul:
One more question @mikealfare, as the issue mentions this flag can't be set via ALTER, only CREATE. Could you point me to where these distinctions are encoded?
> Lifecycle : default except Refresh is bypassed if auto_refresh
on_configuration_change.apply : enable_refresh/refresh_interval_minutes/max_staleness via [ALTER](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_materialized_view_set_options_statement), rest via DROP/[CREATE](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_materialized_view_statement)

mikealfare:
> One more question @mikealfare, as the issue mentions this flag can't be set via ALTER, only CREATE. Could you point me to where these distinctions are encoded?
> 
> > Lifecycle : default except Refresh is bypassed if auto_refresh
> > on_configuration_change.apply : enable_refresh/refresh_interval_minutes/max_staleness via [ALTER](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_materialized_view_set_options_statement), rest via DROP/[CREATE](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_materialized_view_statement)

Good catch, I missed this. This complicates things. This basically means if the value changes, you'll need a full refresh to re-create the materialized view. There's a change object that has a `requires_full_refresh` calculated property here: https://github.com/dbt-labs/dbt-bigquery/blob/bf30b66bab0e96dd4d100f45c2818daa5dde456d/dbt/adapters/bigquery/relation_configs/_options.py#L153-L158. Prior to this PR, all options changes were able to be changed on existing objects, hence the hard-coded `False`. Now we need to see if that changed. The problem is that this object only has the new options on it. These get set here: https://github.com/dbt-labs/dbt-bigquery/blob/bf30b66bab0e96dd4d100f45c2818daa5dde456d/dbt/adapters/bigquery/relation.py#L80-L84. The approach I would take is to set the action to be `alter` if this new option does not change, or set it to be `create` if it does (in the `relation` file). Then condition on the action in the calculated property in `requires_full_refresh` to determine if a full refresh is necessary. Beyond that, I'm not sure what will happen if you submit an alter statement with this setting, even if it's the same. So you might need to pop off this option before passing it to the alter template (but not before passing it to the create template).

mikealfare:
Also, given that this option acts differently and should generate a full refresh, we should add a test for that. You could see how I did that here: https://github.com/dbt-labs/dbt-bigquery/blob/main/tests/functional/adapter/materialized_view_tests/test_materialized_view_cluster_changes.py. Basically swap out the clustering-related pieces for the new setting.

bnaul:
Seems doable, one other thing I'm hitting is that max_staleness is currently ignored (always parsed as None), it's present on the Python table object but as a JSON string representation that can convert to a relativedelta. The standard way a user would define it though would just be like INTERVAL 60 MINUTES, not 0-0 60:0:0. The interval literal rules seem a bit complicated https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#interval_type, is there anywhere that handles this kind of thing already or would comparing user- and BQ API-provided intervals be a new addition here?

mikealfare:
> Seems doable, one other thing I'm hitting is that max_staleness is currently ignored (always parsed as None), it's present on the Python table object but as a JSON string representation that can convert to a relativedelta. The standard way a user would define it though would just be like INTERVAL 60 MINUTES, not 0-0 60:0:0. The interval literal rules seem a bit complicated https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#interval_type, is there anywhere that handles this kind of thing already or would comparing user- and BQ API-provided intervals be a new addition here?

No, there's nowhere that handles this, and I believe the issue you identified (the complexity) is why we moved forward while excluding it from the comparison. I haven't finished the docs for 1.7 yet, but that's certainly a caveat. The workaround is to use `--full-refresh`.

jess-gf:
Hi guys, 
Thank you so much for this pull request.
Do you have any updates if this change is going to be merged soon? =)
Thanks

dbt/adapters/bigquery/relation.py
```diff
@@ -90,8 +90,16 @@ def materialized_view_config_changeset(
         new_materialized_view = cls.materialized_view_from_relation_config(relation_config)
 
         if new_materialized_view.options != existing_materialized_view.options:
+            # allow_non_incremental_definition cannot be changed via ALTER, must recreate
+            if (
```
mikealfare, line 94: I would swap the order here. I think it reads more naturally in the form of "if this setting changed, then recreate, otherwise alter", similar to your comment above.

dbt/adapters/bigquery/relation_configs/_options.py
```diff
@@ -22,6 +23,7 @@ class BigQueryOptionsConfig(BigQueryBaseRelationConfig):
     refresh_interval_minutes: Optional[float] = 30
     expiration_timestamp: Optional[datetime] = None
     max_staleness: Optional[str] = None
+    allow_non_incremental_definition: Optional[bool] = False
```
mikealfare, line 26: This should be `None`, so that it drops out of the config dict if it's not set. It also looks like there is no default according to the [docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#materialized_view_option_list). It's an odd way of saying it, but they say `allow_non_incremental_definition=true is disabled by default` versus `Default: true` as appears elsewhere.

dbt/adapters/bigquery/relation_configs/_options.py
```diff
@@ -158,4 +168,4 @@ class BigQueryOptionsConfigChange(RelationConfigChange):
 
     @property
     def requires_full_refresh(self) -> bool:
-        return False
+        return self.action != RelationConfigChangeAction.alter
```
mikealfare, line 171: I would swap this one too, e.g. `self.action == RelationConfigChangeAction.create`.

tests/functional/adapter/materialized_view_tests/_files.py
```diff
@@ -36,7 +36,8 @@
     cluster_by=["id", "value"],
     enable_refresh=True,
     refresh_interval_minutes=60,
-    max_staleness="INTERVAL 45 MINUTE"
+    max_staleness="INTERVAL 45 MINUTE",
+    allow_non_incremental_definition=True
```
mikealfare, line 40: It makes sense to add this here. But let's add a check for it as well [here](https://github.com/dbt-labs/dbt-bigquery/blob/063baeb62e59210ad1c278f66c752c510bebcb9b/tests/functional/adapter/materialized_view_tests/test_materialized_view_changes.py#L16-L25). It doesn't test the change, we're doing that with other settings, but it at least tests that we're setting it when it's available.

mikealfare (commented):
Thanks for the contribution @bnaul! This looks mostly complete. I had a few nits and few small changes that I'd like to make before merging this. Let me know if you have any concerns with my responses.

mikealfare:
This PR will be continued in https://github.com/dbt-labs/dbt-bigquery/pull/1255. I'm closing this to avoid confusion, however the work will continue. Please refer to #1255 moving forward.

PR #1014, opened 2023-11-09
Fix Broken Python Models

### LINKED ISSUE #1006: [ADAP-1016] [Regression] [1.7] python models don't work

### Current Behavior

all Python models fail to compile with the following compilation error.

`sequence item 2: expected str instance, NoneType found`

[thread from #db-bigquery community Slack](https://getdbt.slack.com/archives/C99SNSRTK/p1699010277788499)

```py
21:23:46  Compilation Error in model thing (models/thing.py)
  sequence item 2: expected str instance, NoneType found
  
  > in macro materialization_table_bigquery (macros/materializations/table.sql)
  > called by model thing (models/thing.py)
```

the model should run

### Steps To Reproduce

1. create the below file in jaffle_shop
1. `dbt seed`
2. `dbt run -s +thing`

weirdly `dbt compile -s thing` works without issue

```py
-- thing.py
def model(dbt, session):
    dbt.config(
        submission_method="serverless",
        dataproc_cluster_name="dbt-test-1"
    )

    my_model = dbt.ref("orders")

    return my_model
```

### Relevant log output

_No response_

### Environment

```markdown
- OS: macOS 13.5
- Python: `3.10.8`
- dbt-core (working version): `1.6.7`
- dbt-bigquery (working version): `1.6.8`
- dbt-core (regression version): `1.7.1`
- dbt-bigquery (regression version): `1.7.0`
```

### Additional Context

perhaps related to #681?

colin-rogers-dbt:
resolves #1006 
Also separates out python model integration tests so they can run in parallel to the rest. 

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/python_submissions.py
+++ b/dbt/adapters/bigquery/python_submissions.py
@@ -1,5 +1,7 @@
 from typing import Dict, Union
 
+from dbt.events import AdapterLogger
+
 from dbt.adapters.base import PythonJobHelper
 from google.api_core.future.polling import POLLING_PREDICATE
 
@@ -17,6 +19,7 @@ from dbt.adapters.bigquery.dataproc.batch import (
 )
 
 OPERATION_RETRY_TIME = 10
+logger = AdapterLogger("BigQuery")
 
 
 class BaseDataProcHelper(PythonJobHelper):
@@ -122,10 +125,18 @@ class ServerlessDataProcHelper(BaseDataProcHelper):
         )
 
     def _get_batch_id(self) -> str:
-        return self.parsed_model["config"].get("batch_id")
+        model = self.parsed_model
+        if "batch_id" in model["config"]:
+            batch_id = model["config"]["batch_id"]
+        else:
+            batch_id = model["unique_id"].replace(".", "-").replace("_", "-") + str(
+                int(model["created_at"])
+            )
+        return batch_id
 
     def _submit_dataproc_job(self) -> Batch:
         batch_id = self._get_batch_id()
+        logger.info(f"Submitting batch job with id: {batch_id}")
         request = create_batch_request(
             batch=self._configure_batch(),
             batch_id=batch_id,
--- a/dbt/adapters/bigquery/python_submissions.py
+++ b/dbt/adapters/bigquery/python_submissions.py
@@ -126,13 +126,9 @@ class ServerlessDataProcHelper(BaseDataProcHelper):
 
     def _get_batch_id(self) -> str:
         model = self.parsed_model
-        if "batch_id" in model["config"]:
-            batch_id = model["config"]["batch_id"]
-        else:
-            batch_id = model["unique_id"].replace(".", "-").replace("_", "-") + str(
-                int(model["created_at"])
-            )
-        return batch_id
+        default_batch_id = model["unique_id"].replace(".", "-").replace("_", "-")
+        default_batch_id += str(int(model["created_at"]))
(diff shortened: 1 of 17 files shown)


dbt/adapters/bigquery/python_submissions.py
```diff
@@ -122,10 +125,18 @@ def _get_job_client(self) -> dataproc_v1.BatchControllerClient:
         )
 
     def _get_batch_id(self) -> str:
-        return self.parsed_model["config"].get("batch_id")
+        model = self.parsed_model
```
mikealfare, line 128: I think this would read better as the default for `get`, in other words:

```python
    def _get_batch_id(self) -> str:
        model = self.parsed_model
        default_batch_id = model["unique_id"].replace(".", "-").replace("_", "-") + str(
            int(model["created_at"])
        )
        return model["config"].get("batch_id", default_batch_id)
```

dbt/include/bigquery/macros/materializations/table.sql
```diff
@@ -110,14 +110,16 @@ df.write \
   .mode("overwrite") \
   .format("bigquery") \
   .option("writeMethod", "indirect").option("writeDisposition", 'WRITE_TRUNCATE') \
+  {%- if partition_config is not none %}
   {%- if partition_config.data_type | lower in ('date','timestamp','datetime') %}
   .option("partitionField", "{{- partition_config.field -}}") \
   {%- if partition_config.granularity is not none %}
-  .option("partitionType", "{{- partition_config.granularity -}}") \
+  .option("partitionType", "{{- partition_config.granularity|upper -}}") \
```
mikealfare, line 117: If we're space-padding the pipe on line 122 (and like it is on line 114), then I would do it here too.

tests/functional/adapter/test_python_model.py
```diff
@@ -29,7 +29,7 @@ def models(self):
         return {"30_sec_python_model.py": blocks_for_thirty_sec}
 
     def test_model_times_out(self, project):
-        result, output = run_dbt_and_capture(["run"], expect_pass=False)
+        result, output = run_dbt_and_capture(["run", "--debug"], expect_pass=False)
```
mikealfare, line 32: I'm pretty sure `"--debug"` needs to be the first item in the list. I ran into issues when I placed it anywhere else because the logs didn't get produced. I'm actually kind of surprised this test passed since you're checking the logs on line 34.

tests/functional/adapter/test_python_model.py
```diff
@@ -216,6 +216,7 @@ def model(dbt, spark):
 """
 
 
+@pytest.mark.skip(reason="Currently failing as run_started_at is the same across dbt runs")
```
mikealfare, line 219: To make sure I understand the comment, is run_started_at referring to `model["created_at"]` above? In other words the two `batch_id`s are the same?

colin-rogers-dbt: No the fixture for this test specifies the batch_id as:
```batch_id: '{{ run_started_at.strftime("%Y-%m-%d-%H-%M-%S") }}-python-array'```
`run_started_at` appears to be the same for both runs (I'm guessing because of how we setup our integ test runs). We need to come up with a way to specify a different batch_id for different runs

tests/functional/adapter/test_python_model.py
colin-rogers-dbt: This got a left in by accident

mikealfare (approved):

PR #1042, opened 2023-12-05
retry wait for result independently from job creation

### LINKED ISSUE #1045: [ADAP-1062] [Bug] Retries on wait for result step is recreating the whole job

### Current Behavior

Retries on wait for result step is recreating the whole BigQuery job which means it's duplicating the run.

### Expected Behavior

If a job has been started successfully, we should retry to connect to the actual job status and not create a new one.

### Steps To Reproduce

This bug is hard to reproduce as it depends on potential Google's being in error or the network to fail on the request.
It might be mimicked by changing the code the code return an error instead of waiting for the result.

### Relevant log output

```shell
2023-12-02T04:43:31.713+01:00	[0m03:43:31 BigQuery adapter: https://console.cloud.google.com/bigquery?project=prodcution&j=bq:US:11111111-2222-3333-4444-aaaaaaaaaaaa&page=queryresults

... 1,25 hour latter

2023-12-02T05:55:12.921+01:00	[0m04:55:12 BigQuery adapter: Retry attempt 1 of 3 after error: ConnectionError(ProtocolError('Connection aborted.', RemoteDisconnected('Remote end closed connection without response')))

2023-12-02T05:55:12.922+01:00	[0m04:55:12 BigQuery adapter: Reopening connection after ConnectionError(ProtocolError('Connection aborted.', RemoteDisconnected('Remote end closed connection without response')))

2023-12-02T05:55:14.347+01:00	[0m04:55:14 BigQuery adapter: https://console.cloud.google.com/bigquery?project=teads-prod-analytics-etl&j=bq:US:11111111-2222-3333-4444-bbbbbbbbbbbb&page=queryresults
```

### Environment

```markdown
- OS: Ubuntu 20.04
- Python: 3.10
- dbt-core: 1.7.0
- dbt-bigquery: 1.7.0
```

### Additional Context

_No response_

github-christophe-oudar:
resolves #1045

### Problem

When an error occurs either on job creation or waiting for the result, the job creation + wait result step is retried.
Then the underlying wait for result step might "fail" (as it's polling for the result every X seconds) and a network error... can lead to retry the whole job.
If the job isn't idempotent => it leads to a bug (what happened for a coworker).
if the job is idempotent => you likely wasted slot time/BQ resources.

### Solution

To solve that, let's split the step in 2 functions that are both retried on their own so that we retry accessing the running job.

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -492,17 +492,23 @@ class BigQueryConnectionManager(BaseConnectionManager):
         job_creation_timeout = self.get_job_creation_timeout_seconds(conn)
         job_execution_timeout = self.get_job_execution_timeout_seconds(conn)
 
-        def fn():
+        def job_creation_fn():
             return self._query_and_results(
-                client,
-                sql,
-                job_params,
-                job_creation_timeout=job_creation_timeout,
+                client, sql, job_params, job_creation_timeout=job_creation_timeout
+            )
+
+        query_job = self._retry_and_handle(msg=sql, conn=conn, fn=job_creation_fn)
+
+        def wait_for_job_results_fn():
+            return self._wait_for_results(
+                query_job=query_job,
                 job_execution_timeout=job_execution_timeout,
                 limit=limit,
             )
 
-        query_job, iterator = self._retry_and_handle(msg=sql, conn=conn, fn=fn)
+        query_job, iterator = self._retry_and_handle(
+            msg=sql, conn=conn, fn=wait_for_job_results_fn
+        )
 
         return query_job, iterator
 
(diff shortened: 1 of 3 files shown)


dbt/adapters/bigquery/connections.py
```diff
@@ -787,7 +789,7 @@ def reopen_conn_on_error(error):
                 target=fn,
                 predicate=_ErrorCounter(self.get_job_retries(conn)).count_error,
                 sleep_generator=self._retry_generator(),
-                deadline=self.get_job_retry_deadline_seconds(conn),
+                timeout=self.get_job_retry_deadline_seconds(conn),
```
github-christophe-oudar, line 792: it's just that the previous field is deprecated in the driver, it's the same behavior

McKnight-42:
@github-christophe-oudar wanted to point you to this open PR I have as I think these are related https://github.com/dbt-labs/dbt-bigquery/pull/977

McKnight-42:
@github-christophe-oudar The ticket is still being worked on. I did have to set it aside for a bit due to some other work and traveling but it's on my board. I plan to set some time to dig into these pr's and problems a little more over the next few days and will keep you updated.

github-christophe-oudar:
Ok, great to know!
I think both approaches could be combined.
The sooner we can deal with that bug, the better 🙏

PR #1060, opened 2023-12-18
Implementation of metadata-based freshness

### LINKED ISSUE #938: [ADAP-912] Support Metadata Freshness
### Describe the feature

Support metadata-based freshness by implementing the new macro and feature flag described in https://github.com/dbt-labs/dbt-core/issues/7012.

### Who will this benefit?

Everyone who wants faster freshness results from BigQuery!

mikealfare:
resolves #938

### Problem

The current implementation of source freshness requires querying the data and requires the user provide a datetime field. This is slower and more expensive than it should be. It doesn't scale across multiple models. And some models do not have an appropriate datetime field.

### Solution

Use source metadata where available.

### Checklist

### DIFF
--- a/.changes/unreleased/Features-20231218-155409.yaml
+++ b/.changes/unreleased/Features-20231218-155409.yaml
@@ -1,0 +1,6 @@
+kind: Features
+body: Add support for checking table-last-modified by metadata
+time: 2023-12-18T15:54:09.69635-05:00
+custom:
+  Author: mikealfare
+  Issue: "938"


tests/boundary/test_bigquery_sdk.py
mikealfare, line 1: I added this test because we're seeing odd behavior with the `get_table` method (a BQ Client method). If we run `get_table` during source freshness (line 726 in `impl.py`) on a table that does not exist, it hangs without returning the expected `NotFound` error, even when providing `retry`. However when we do the same here, it seems to behave as expected.

colin-rogers-dbt: do we understand this well enough to raise a bug with bigquery? https://github.com/googleapis/python-bigquery/issues

colin-rogers-dbt: There's this issue: https://github.com/googleapis/python-bigquery/issues/1674

mikealfare:
Removing `backport 1.7.latest` as this has actually been merged into `1.7.latest` already.

mikealfare:
This PR reflects the changes that were merged directly into 1.7.latest via https://github.com/dbt-labs/dbt-bigquery/pull/1072.

colin-rogers-dbt (approved):

PR #1089, opened 2024-02-05
Create BQ version of string_literal with correct quoting

### LINKED ISSUE #1088: [Bug] For BigQuery, a multiline string needs to be escaped with 3 quotes instead of 1 like in `default__string_literal()`

### Current Behavior

`dbt.string_literal(val)` is defined in `dbt-adapters` but not overwritten in `dbt-bigquery`.

Calling `dbt.string_literal(val)` fails in BigQuery when `val` contains newline characters. This is because strings with newlines characters need to be "triple-escaped" in BigQuery ([docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#literals))

### Expected Behavior

Calling `dbt.string_literal(val)` in BigQuery should work with a `val` containing newlines

### Steps To Reproduce

Create a model with 

```
select {{ dbt.string_literal('my multiline 
string') }} as test
```

### Relevant log output

_No response_

### Environment

_No response_

### Additional Context

_No response_

b-per:
resolves #1088

### Problem

`dbt.string_literal(val)` doesn't work in BQ when the `val` contains newlines

### Solution

We need a BQ dispatched version of the macro with "triple quoting", like described [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#literals).

### Checklist


mikealfare:
Thanks for the PR! Mind throwing a test on here as well to verify it's working and ensure it continues to work? I think a successful execution of a simple model with the use case in your issue would suffice.

colin-rogers-dbt:
I think this is already covered by a test and it's failing: https://github.com/dbt-labs/dbt-bigquery/actions/runs/7919387344/job/21620044225?pr=1089#step:7:1033

mikealfare:
> I think this is already covered by a test and it's failing: https://github.com/dbt-labs/dbt-bigquery/actions/runs/7919387344/job/21620044225?pr=1089#step:7:1033

This looks like one of the new tests for the unit testing framework. We need to handle the geometry and geography data types differently, specifically within `cast` and `safe_cast`. Perhaps it wasn't implemented completely for `dbt-snowflake`. Either way, I think it's unrelated to this change.

mikealfare (approved):

PR #1118, opened 2024-02-22
Point to feature branch of `dbt-tests-adapter` for testing, this should not be merged

mikealfare:


mikealfare:
We can avoid PRs like this in the future with this ticket: https://github.com/dbt-labs/dbt-adapters/issues/101

PR #1140, opened 2024-03-18
Add Bigquery internal build process workflow 

VersusFacit:
resolves 38 of epic

This PR adds a workflow file for bigquery testing. It is based on [the already working redshift one](https://github.com/dbt-labs/dbt-redshift/pull/732). I assume there will be some quirks to work out. Also, we'll need to add aws secrets.


emmyoop (changes requested):
All the same questions/changes in https://github.com/dbt-labs/dbt-redshift/pull/732

emmyoop (approved):
Approving as is for now with the understanding that more PRs will follow on after testing since none of this can be tested before the workflow gets merged.  Since this is a workflow that will be manually submitted there is little risk in this approach.

PR #1205, opened 2024-04-26
Add configuration options for `enable_list_inference` and `intermediate_format` for python models

### LINKED ISSUE #1047: [ADAP-1063] [Bug] BigQueryException: Error while reading data, error message: Schema mismatch

### Current Behavior

Dataproc job fails unexpectedly using dbt-core>1.7.0 and dbt-bigquery>1.7.0 but works using dbt-core<1.6.9 and dbt-bigquery<1.6.9 with schema mismatch error.
 

### Expected Behavior

Updating dbt should not result in dataproc job failing

### Steps To Reproduce

The column type is:
```
column_name      | RECORD | REPEATED |
  - my_list             | STRING | REPEATED | 
  - name                | STRING | NULLABLE
```

### Relevant log output

```shell
Caused by: com.google.cloud.spark.bigquery.repackaged.com.google.cloud.bigquery.BigQueryException: Error while reading data, error message: Schema mismatch: referenced variable 'column_name.list.element.my_list.list.element' has array levels of 1, while the corresponding field path to Parquet column has 0 repeated fields File: gs://<my-bucket>/.spark-bigquery-application_1697116734723_1013-4f752a25-0f69-441e-8641-54bee7fcb8ce/part-00007-ab709b08-3bd5-481d-a6d4-f77989c194f8-c000.snappy.parquet
```

### Environment

```markdown
- OS: macos
- Python: 3.11
- dbt-core: >1.7.0
- dbt-bigquery: >1.7.0
```

### Additional Context

Would impact anyone moving from Python models version 1.6 to 1.7 using BQ

Currently can't reproduce, will pair with Doug to repro.

_No response_

### LINKED ISSUE #1114: [Bug] List types don't save to BigQuery using 1.7 (but did in 1.5)

### Current Behavior

Using the following code:
```python
import pandas as pd

def model(dbt, session):
    dbt.config(
        submission_method="cluster",
        dataproc_cluster_name="dbt-python",
        packages=["pandas", "numpy"]
    )

    # Define schema for the DataFrame
    schema = "id INT, embedding ARRAY<INT>"

    # Create a DataFrame
    data = [(5, [1, 2, 3]), (6, [4, 5, 6]), (7, [7, 8, 9])]
    df = session.createDataFrame(data, schema=schema)

    return df
```

The database result is:
image (7)

As you can see the `embedding` column has no data

### Expected Behavior

Running the same Python code:
```
import pandas as pd

def model(dbt, session):
    dbt.config(
        submission_method="cluster",
        dataproc_cluster_name="dbt-python",
        packages=["pandas", "numpy"]
    )

    # Define schema for the DataFrame
    schema = "id INT, embedding ARRAY<INT>"

    # Create a DataFrame
    data = [(5, [1, 2, 3]), (6, [4, 5, 6]), (7, [7, 8, 9])]
    df = session.createDataFrame(data, schema=schema)

    return df
```

Should result in the following result in the target table in the data warehouse:
image (8)

### Steps To Reproduce

NOTE: This worked without issue in dbt 1.5.9 and loaded the data correctly.

1. Using dbt-core 1.7.8 with BigQuery adapter 1.7.4 running with a DataProc cluster in GCP
2. Create a new dbt python model using the code described in the behaviour section
3. Run the model
4. Observe the results in the target table

### Relevant log output

_No response_

### Environment

```markdown
- OS: Mac OS 14 (Dataproc OS - 2.1.40-debian11)
- Python: 3.9.9
- dbt: 1.7.8
- bigquery-connector-version: 1.2.0
- spark-bigquery-connector-version: 0.36.1
```

bigquery

### Additional Context

As stated previously, this worked without issue on dbt 1.5.9 and loaded the data as expected. After updating to dbt 1.7.8 this has now broken and the data is no longer loaded correctly.

mikealfare:
resolves #1047 resolves #1114
docs https://github.com/dbt-labs/docs.getdbt.com/issues/5375

### Problem

See the attached issues.

### Solution

- default enableListInference to true
- add configuration option to turn enableListInference off
- add configuration option for intermediateFormat which is another approach to solve the problem

### Checklist

- [ ] The docs PR on docs.getdbt.com is ready for review


dbt/adapters/bigquery/impl.py
mikealfare, line 1: @dataders / @Fleid These are two new configs that I'm adding for python models. They correspond to `spark` configs of `enableListInference` and `intermediateFormat`. Are you comfortable adding these in? Would you change anything about them?

dbt/include/bigquery/macros/materializations/table.sql
```diff
@@ -49,12 +49,19 @@
 from pyspark.sql import SparkSession
 {%- set raw_partition_by = config.get('partition_by', none) -%}
 {%- set raw_cluster_by = config.get('cluster_by', none) -%}
+{%- set enable_list_inference = config.get('enable_list_inference', true) -%}
```
mikealfare, line 52: Since this spawned from a regression, setting the default to `true` effectively mitigates the regression.

tests/functional/python_model_tests/test_list_inference.py
mikealfare, line 1: I ran these multiple times. It looks like what's happening is that the tests will always pass when run in series, but when run in parallel a few usually fail. All I can think of is that the session parameters that get set in one model are affecting the other models. Perhaps this setting can only be at the "run" level?

dbt/adapters/bigquery/impl.py
Fleid: Hey @mikealfare do you know why these parameters are now necessary?
Do we have the root cause on the regression here? Did anything change on BQ?

I'm not against adding these, I'm just wary of adding stuff not knowing why we need them now.

mikealfare: I think the issue is that we went from `"writeMethod", "direct"` to `"writeMethod", "indirect"` [here](https://github.com/dbt-labs/dbt-bigquery/blob/461fa04abec0129304c2cde414624f051789f5a2/dbt/include/bigquery/macros/materializations/table.sql#L112).

While we accounted for some of the side effects, we didn't account for all of them. I did some research and it looks like it's being passed through to `spark-bigquery-connector`. There's a write-up [here](https://github.com/GoogleCloudDataproc/spark-bigquery-connector#writing-data-to-bigquery) discussing the difference between `direct` and `indirect`.

Ultimately, I think the issue is that writing (indirectly) to parquet (the default I believe) turns out to have more nuance than writing directly to a BQ table.

mikealfare: I thought about this more and realized that we can also allow `writeMethod` to be configurable. If we add that as a third configurable input, we could get back to Scenario A as the default. However, that would break folks who opted in to partitioning (Scenario B). Let me know what you think and I can update this as needed.

tests/functional/python_model_tests/test_list_inference.py
mikealfare: I moved them to the python model set of tests and now it passes. I think they need to be run in series and not parallel.

dbt/adapters/bigquery/impl.py
Fleid: Let's go for scenario C then!

How you are putting it, I don't feel the need yet to expose `writeMethod`. But we should go there if we get any more trouble from these settings.

tests/functional/python_model_tests/files.py
```diff
@@ -0,0 +1,148 @@
...
+            {"column_name": [{"name": "hello", "my_list": ["h", "e", "l", "l", "o"]}]},
+        ]
+    )
+
+    return df
+"""
+
+
+# this should fail
+DISABLE_LIST_INFERENCE = """
```
VersusFacit, line 84: Unused. Did you want a test for testing that this fails?

tests/functional/python_model_tests/test_list_inference.py
```diff
@@ -0,0 +1,36 @@
+"""
+This test case addresses this regression: https://github.com/dbt-labs/dbt-bigquery/issues/1047
+
+As the comments point out, the issue appears to be that the default settings are:
```
VersusFacit, line 4: this comment is true but won't be post merge. Maybe rephrase to say "this issue appears if/when"

VersusFacit (commented):
Love this PR. Small comments

tests/functional/python_model_tests/files.py
mikealfare: I originally had one that showed it was failing, but I don't think we need to confirm that it will continue to fail. I removed it from the test, but forgot to remove it from this file.

tests/functional/python_model_tests/test_list_inference.py
mikealfare: I updated the comment as suggested, and made a few other updates since this comment went out of date with what was done below (e.g. `files` instead of `models`).

VersusFacit (approved):

PR #1213, opened 2024-05-01
Remove usage of `__TABLES__`

### LINKED ISSUE #113: [CT-162] Upgrade from the __tables__ construct to the information_schema.tables construct

mikealfare:
resolves #113

rebase of #364


dbt/include/bigquery/macros/catalog/catalog.sql
```diff
@@ -1,31 +1,31 @@
...
-            when materialized_views.table_name is not null then 'materialized view'
-            when tables.type = 1 then 'table'
-            when tables.type = 2 then 'view'
+        tables.table_catalog,
+        tables.table_schema,
+        coalesce(REGEXP_EXTRACT(tables.table_name, '^(.+)[0-9]{8}$'), tables.table_name) as table_name,
+        tables.table_name as shard_name,
+        REGEXP_EXTRACT(tables.table_name, '^.+([0-9]{8})$') as shard_index,
+        REGEXP_CONTAINS(tables.table_name, '^.+[0-9]{8}$') and tables.table_type = 'BASE TABLE' as is_date_shard,
+        case tables.table_type
```
mikealfare, line 9: Seeds are showing up as `external` now, otherwise this seems to be working.

dbeatty10:
Key comments from description of https://github.com/dbt-labs/dbt-bigquery/pull/364 that we'll want to consider sufficiently prior to merging this PR.

### Trade-offs and outstanding questions

The problem is that it doesn't contain `row_count` or `size_bytes`, so it is not a perfect drop-in replacement. This could be possibly overcome by using the [`INFORMATION_SCHEMA.TABLE_STORAGE`](https://cloud.google.com/bigquery/docs/information-schema-table-storage) view to get the number of rows + the size in logical bytes, but that has its own complications (see below).

1. Is it okay to go from reporting the row count and size in bytes and then degrade to `0` for both?
    - Is there a way to make [this](https://github.com/GeneralMills/gmi_common_dbt_utils/pull/5/files#diff-b8a2cba11fa5e23d49a5af4c4f49876abde8078152a1dcdba7454846ebec59f0R46-R52) work without an environment variable? e.g., is there a way for the adapter to know the relevant region(s)?
1. Are there any region-specific effects that would could cause a table/view/external to not be reflected?
1. Is this method slower, faster, or the same?

amychen1776:
Given the tradeoffs, I'm quite weary of removing displaying row count and size given how many folks have found that to be helpful.

dbt/include/bigquery/macros/catalog/catalog.sql
```diff
@@ -1,31 +1,31 @@
...
+        tables.table_type = 'BASE TABLE' as is_table,
         JSON_VALUE(table_description.option_value) as table_comment,
-        tables.size_bytes,
-        tables.row_count
-    from {{ information_schema.replace(information_schema_view='__TABLES__') }} tables
-    left join {{ information_schema.replace(information_schema_view='MATERIALIZED_VIEWS') }} materialized_views
-        on materialized_views.table_catalog = tables.project_id
-        and materialized_views.table_schema = tables.dataset_id
-        and materialized_views.table_name = tables.table_id
+        partitions.total_logical_bytes as size_bytes,
```
mikealfare, line 17: @dbeatty10 @amychen1776 It looks like you're saying that we're not getting the size nor the row count. Are these two fields (`partitions.total_logical_bytes` and `partitions.total_rows`) coming through as 0?

dbeatty10: I haven't tried out this PR, so I'm not sure how `partitions.total_logical_bytes` and `partitions.total_rows` are coming through. How are they coming through for you @mikealfare ?

It would be awesome if they successfully preserve non-null and non-zero values for `size_bytes` and `total_rows`!

mikealfare: Sounds like we need an integration test or two. I know I worked on this when I thought this was the source of a bug a few weeks ago. It turned out to not be the case though, so I left it there. @jtcohen6 @amychen1776, would you consider prioritizing this for the upcoming sprint so that we could capture the value here? We would get two main things:
- get off of `__TABLES__`, which is an undocumented dependency (we should be using `TABLES`
- reduce permissions required to run `dbt-bigquery`

PR #1255, opened 2024-06-11
Add support for non-incremental materialized views

### LINKED ISSUE #589: [ADAP-464] Materialized Views - Non-incremental support
As [commented in the core MV thread](https://github.com/dbt-labs/dbt-core/issues/6911#issuecomment-1503925035), BigQuery is adding support for [non-incremental MVs](https://cloud.google.com/bigquery/docs/materialized-views-create#non-incremental).

From our side, it means that we have an additional setting to support for MVs in BQ: `allow_non_incremental_definition`: `true`/`false`. This setting is covered by `on_configuration_change` via DROP/CREATE since it can't be changed via ALTER.

When set, it requires `max_staleness` to be set. I'd rather leave BQ errors out here for now - this is a preview feature and I'm not sure this won't change before GA. We should not check this.

___
#### BigQuery

[BigQuery Materialized Views](https://cloud.google.com/bigquery/docs/materialized-views-intro)

- Configurations:
  - enable_refresh : BOOLEAN, optional (database level [default](https://cloud.google.com/bigquery/docs/materialized-views-manage#automatic_refresh) true) (see lifecycle for effect)
  - refresh_interval_minutes : integer, optional (database level [default](https://cloud.google.com/bigquery/docs/materialized-views-manage#automatic_refresh) at 30), only valid for enable_refresh true or default
  - max_staleness : time before bypassing the MV and running the statement instead (in preview but I like it!)
  - **allow_non_incremental_definition : BOOLEAN, optional (database default `false`)**
  - From models: [Partition / Cluster](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#using-table-partitioning-and-clustering), [TTL](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#controlling-table-expiration), [Labels](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#labels-and-tags)  
- Lifecycle : _default_ except Refresh is bypassed if auto_refresh
  - `on_configuration_change`.`apply` : enable_refresh/refresh_interval_minutes/max_staleness via [ALTER](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_materialized_view_set_options_statement), rest via DROP/[CREATE](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_materialized_view_statement)

### LINKED ISSUE #672: Materialized Views - Non-incremental support

mikealfare:
resolves dbt-labs/dbt-adapters#589

This PR continues @bnaul's work in https://github.com/dbt-labs/dbt-bigquery/pull/1011.

This PR has become more complicated than expected and will require some discussion prior to moving forward. There are a number of items that need to be addressed:

- align on approach for managing options directly on the MV
- align on approach of templating each option explicitly in jinja
- update tests to catch all options
- update change tests to test all (reasonably) change scenarios
- update the macros that swap MVs for views/tables with MVs as certain options do not pertain to views/tables and cause issues when they appear in the model

### Checklist


dbt/adapters/bigquery/connections.py
```diff
@@ -320,7 +320,7 @@ def format_rows_number(self, rows_number):
         return f"{rows_number:3.1f}{unit}".strip()
 
     @classmethod
-    def get_google_credentials(cls, profile_credentials) -> GoogleCredentials:
+    def get_google_credentials(cls, profile_credentials) -> GoogleCredentials.Credentials:
```
mikealfare, line 323: This was failing `mypy`. I'm not sure how it wouldn't have failed in the past.

mikealfare:
Closing and re-opening to re-trigger CI post GH outage.

tests/functional/adapter/materialized_view_tests/_mixin.py
```diff
@@ -46,7 +46,7 @@ def my_other_base_table(self, project) -> BaseRelation:
         )
 
     @pytest.fixture(scope="function", autouse=True)
-    def setup(self, project, my_base_table, my_other_base_table, my_materialized_view):  # type: ignore
+    def setup(self, project, my_materialized_view):
```
mikealfare, line 49: These arguments do not get used, and were causing typing issues.

dbt/adapters/bigquery/connections.py
VersusFacit: I think they changed things recently because I _believe_ this is what caused the error on 1.3. Must have been an upstream spec change because our code didn't on 1.3. Good catch on the solution! Module vs. type

dbt/adapters/bigquery/impl.py
```diff
@@ -115,6 +115,7 @@ class BigqueryConfig(AdapterConfig):
     max_staleness: Optional[str] = None
     enable_list_inference: Optional[bool] = None
     intermediate_format: Optional[str] = None
+    allow_non_incremental_definition: Optional[bool] = None
```
VersusFacit, line 118: Should this be a default of False? Having `None` introduces a Falsey third state, and I'm not sure it's needed

dbt/adapters/bigquery/relation.py
```diff
@@ -90,9 +90,10 @@ def materialized_view_config_changeset(
         new_materialized_view = cls.materialized_view_from_relation_config(relation_config)
 
         if new_materialized_view.options != existing_materialized_view.options:
-            config_change_collection.options = BigQueryOptionsConfigChange(
-                action=RelationConfigChangeAction.alter,
-                context=new_materialized_view.options,
+            # get an options change object with only the options that have changed
```
VersusFacit, line 93: As opposed to what. Pardon, just not sure what this comment refers to with the prexisting code. Were we instantiating the object in a different way? I see you're no longer using kwargs which is fine, but what's the intent here?

dbt/adapters/bigquery/relation_configs/_options.py
```diff
@@ -22,6 +23,7 @@ class BigQueryOptionsConfig(BigQueryBaseRelationConfig):
     refresh_interval_minutes: Optional[float] = 30
     expiration_timestamp: Optional[datetime] = None
     max_staleness: Optional[str] = None
+    allow_non_incremental_definition: Optional[bool] = None
```
VersusFacit, line 26: Similar comment to above

dbt/adapters/bigquery/relation_configs/_options.py
```diff
@@ -158,4 +176,19 @@ class BigQueryOptionsConfigChange(RelationConfigChange):
...
+    ) -> Self:
+        new_options_dict = asdict(new_options)
+        existing_options_dict = asdict(existing_options)
+        option_diffs_dict = {
+            k: v
+            for k, v in new_options_dict.items()
+            if new_options_dict[k] != existing_options_dict[k]
+        }
+        option_diffs = BigQueryOptionsConfig.from_dict(option_diffs_dict)
+        return cls(action=RelationConfigChangeAction.alter, context=option_diffs)
```
VersusFacit, line 194: Oh I think what you were doing ahead now. I'm not sure I like burrying the alter here. May be a matter of preference. I kind of what to float this action param up for visibility when creating the object as opposed to hardcoding the value down here.

dbt/adapters/bigquery/relation.py
mikealfare: We were creating an options object with all options, not just those that changed, and it was creating issues when altering the options. For example, options that force a full refresh would always show up in the change even if they didn't change, hence changing an option that could be implemented via an `ALTER` statement would still trigger a full refresh due to the presence of the other options.

dbt/adapters/bigquery/impl.py
mikealfare: `None` values get removed from the config entirely whereas `False` or `True` will get set. Also, the BQ "default" is nothing, not `False`. See my comment [here](https://github.com/dbt-labs/dbt-bigquery/pull/1011#discussion_r1631575612).

dbt/adapters/bigquery/relation_configs/_options.py
mikealfare: I don't understand, but I'll comment with what's coming to mind.

Most things are an `alter` in this framework. The other options are `create` and `drop`, which really was only used for indexes in `dbt-postgres` where the entire object (index, not the relation) would be created or dropped instead of altered. To alter an index in `dbt-postgres` you need to issue two statements, one to drop the existing, and another to recreate it with the new config. In this case, we're really just altering the options on the relation (in this case an MV, but could be a table in the future). Sometimes that happens via a full refresh, which gets determined based on the context.

I agree we could spend more time on this framework to make it easier to parse (and I would like to do so if given capacity), but updating that would be outside of the scope of this particular change. Hence I left it with the default `alter` since I need to provide something. Unfortunately, if you follow the code through, this attribute never gets used, only `context` does.

dbt/adapters/bigquery/impl.py
VersusFacit: Great. That's something I'll remember for later with respect to Bigquery. We historically live in a True-False exclusive world but I agree it makes sense to do this here.

dbt/adapters/bigquery/relation.py
VersusFacit: I'd like to indicate that better in the comment. I don't know how it would be good to say that but something like "This is how is used to be done. This is why we were are doing it this way deliberately. This resulted in X. We want to avoid X."

dbt/adapters/bigquery/relation_configs/_options.py
VersusFacit: I agree on not scoping up. I'd just rather have an empty param here and pass in the `alter` deliberately at the function call level above. Right now you're making an assumption that I worry may not be obvious to the next person. But with the generic param passing this argument from the function call, it would be quite straightforward for adding others as needed (even if it never happens, I think it'd read clearer)

dbt/adapters/bigquery/impl.py
colin-rogers-dbt: Maybe a code comment at the top explaining this is a practice used by bigquery with bools in their api design would help, I agree that it's counterintuitive and will likely trip up folks in the future

dbt/adapters/bigquery/impl.py
```diff
@@ -115,6 +115,7 @@ class BigqueryConfig(AdapterConfig):
     max_staleness: Optional[str] = None
```
colin-rogers-dbt, line 115: do we need a post_init check to enforce that this is required if `allow_non_incremental_definition` is also specified?

tests/functional/adapter/materialized_view_tests/test_materialized_view_changes.py
```diff
@@ -63,6 +65,9 @@ def change_config_via_replace(project, materialized_view):
             initial_model.replace(old_partition, new_partition)
             .replace("'my_base_table'", "'my_other_base_table'")
             .replace('cluster_by=["id", "value"]', 'cluster_by="id"')
+            .replace(
```
colin-rogers-dbt, line 68: should we also test going from allow_non_incremental_definition=False to allow_non_incremental_definition=True

dbt/adapters/bigquery/relation_configs/_options.py
mikealfare: When you say to pass it in to the function call above, are you saying to pass it in to `from_options_configs`? While it's static now, it may not be in the future. In fact, I could see accounting for "unsetting" options in the future, would would just be a drop. Then I would need to compare the two options configs to see if it's an alter or a drop/create scenario. That would make `from_options_configs` duplicative. My intent for this method was basically "here are my two options configs, please tell me what I need to do to get from A to B". I don't think supplying the action as an argument as well makes it more readable. I could set the action prior to the return with a comment that all diffs are alter diffs at this point, which may change in the future.

dbt/adapters/bigquery/impl.py
mikealfare: I could add validation rules if desired. There wasn't really an interest previously in doing so, but there's a small framework for running validation rules already, so I can tap into that.

dbt/adapters/bigquery/impl.py
mikealfare: It looks like the practice in general is to default all values for `AdapterConfig` to `None` (likely to make optional entries in the config `dict` via [filter_null_values](https://github.com/dbt-labs/dbt-common/blob/e30f65f6b93eed0422295a4282a892cdfe4b59d5/dbt_common/utils/dict.py#L11-L12)). For example, [here](https://github.com/dbt-labs/dbt-snowflake/blob/efe6df3c2941f722ec3e7449e3c0e0c019c435bf/dbt/adapters/snowflake/impl.py#L26-L36) is the same thing for `dbt-snowflake`. We have `Optional[bool]`s there as well, all defaulted to `None`. We then have a second "defaulting" behavior when we hit `BigQueryOptionsConfig` below. I'll mark this as resolved, that as unresolved, and continue there.

dbt/adapters/bigquery/relation_configs/_options.py
mikealfare: This continues the comment above regarding defaulting `Optional[bool]`s to `None` in `BigQueryConfig`. The intent in this section is to match the defaulting behavior of BigQuery according to their docs. The conclusion that we reached above is that we should have a comment regarding the fact that BQ effectively allows a third falsy value (`None`) for some booleans (`allow_non_incremental_definition`). There is a link in the docstring to those configurations. While I don't want to pull too much information from that link (to avoid getting out of sync with BQ docs), I will add a generic comment to the top that indicates that some booleans are option in BQ in the sense that there is literally no setting if it's not provided (versus defaulting to false). Does that work?

dbt/adapters/bigquery/relation.py
mikealfare: I'll update the comment.

mikealfare: After going around the block on this a few times today, I'm not adopting the approach of simply submitting the entire new options config. This comment no longer applies. However there are new comments for the new scenario. I'm looking for feedback on those comments as a result of this thread.

dbt/adapters/bigquery/relation_configs/_options.py
mikealfare: With the change of providing the entire options config as context for the change, there is no need for a `from_options_config` method. Hence this is now back in the calling method. I think this is what you were asking for, but please let me know if that's incorrect.

yu-iskw:
@mikealfare What should we do in the rest? If there is anything I can help, please let me know.

mikealfare:
This PR has been migrated to dbt-labs/dbt-adapters#823

PR #1260, opened 2024-06-17
update user docs-issue workflow

McKnight-42:

### Problem
We are wanting to make sure all adapters have a `docs-issue` workflow and if it exists currently make sure it is in same state as `dbt-core` workflow to where we trigger off issues now. for parity.

### Solution

- [x] check if docs-issue exists
- [ ] if does update, if doesn't add workflow
- [ ] add user docs label if doesn't exist
### Checklist

### DIFF
--- a/.github/workflows/docs-issue.yml
+++ b/.github/workflows/docs-issue.yml
@@ -1,19 +1,18 @@
 # **what?**
-# Open an issue in docs.getdbt.com when a PR is labeled `user docs`
+# Open an issue in docs.getdbt.com when an issue is labeled `user docs` and closed as completed
 
 # **why?**
 # To reduce barriers for keeping docs up to date
 
 # **when?**
-# When a PR is labeled `user docs` and is merged.  Runs on pull_request_target to run off the workflow already merged,
-# not the workflow that existed on the PR branch.  This allows old PRs to get comments.
+# When an issue is labeled `user docs` and is closed as completed.  Can be labeled before or after the issue is closed.
 
 
-name: Open issues in docs.getdbt.com repo when a PR is labeled
-run-name: "Open an issue in docs.getdbt.com for PR #${{ github.event.pull_request.number }}"
+name: Open issues in docs.getdbt.com repo when an issue is labeled
+run-name: "Open an issue in docs.getdbt.com for issue #${{ github.event.issue.number }}"
 
 on:
-  pull_request_target:
+  issues:
     types: [labeled, closed]
 
 defaults:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-issue.yml
```diff
@@ -1,43 +1,41 @@
...
     # risk of duplicaton of issues being created due to merge and label both triggering this workflow to run and neither having
     # generating the comment before the other runs.  This lives here instead of the shared workflow because this is where we
     # decide if it should run or not.
     if: |
-      (github.event.pull_request.merged == true) &&
-      ((github.event.action == 'closed' && contains( github.event.pull_request.labels.*.name, 'user docs')) ||
-      (github.event.action == 'labeled' && github.event.label.name == 'user docs'))
+      (github.event.issue.state == 'closed' &&
+       github.event.issue.state_reason == 'completed' &&
+       contains( github.event.issue.labels.*.name, 'user docs'))
```
mikealfare, line 35: I think we need to account for the two types here (labeled and closed) as we did for PRs. This looks like just the scenario for closed.

McKnight-42: Does the contains() function not handle that as the add part of the statement? that was my understanding from reading it that we are looking for closed and labeled to kick off the workflow.

.github/workflows/docs-issue.yml
```diff
@@ -1,43 +1,42 @@
...
+    # we only want to run this when the issue is closed as completed and the label `user docs` has been assigned.
+    # If this logic does not exist in this workflow, it runs the
     # risk of duplicaton of issues being created due to merge and label both triggering this workflow to run and neither having
     # generating the comment before the other runs.  This lives here instead of the shared workflow because this is where we
     # decide if it should run or not.
     if: |
-      (github.event.pull_request.merged == true) &&
-      ((github.event.action == 'closed' && contains( github.event.pull_request.labels.*.name, 'user docs')) ||
-      (github.event.action == 'labeled' && github.event.label.name == 'user docs'))
+      (github.event.issue.state == 'closed' &&
```
mikealfare, line 33: We want the issue to be closed and completed for either scenario. Right now the labeled scenario will execute regardless of the state and state_reason. We basically want the issue equivalent of a PR being merged, as it's used in the current workflow.

McKnight-42:
`bigquery / python 3.8 / ubuntu-latest Expected` stays at expected due to not having a change that triggers tests

colin-rogers-dbt (approved):

.github/workflows/docs-issue.yml
```diff
@@ -1,43 +1,41 @@
...
+    # If this logic does not exist in this workflow, it runs the
     # risk of duplicaton of issues being created due to merge and label both triggering this workflow to run and neither having
     # generating the comment before the other runs.  This lives here instead of the shared workflow because this is where we
     # decide if it should run or not.
     if: |
-      (github.event.pull_request.merged == true) &&
-      ((github.event.action == 'closed' && contains( github.event.pull_request.labels.*.name, 'user docs')) ||
+      (github.event.issue.state == 'closed' && github.event.issue.state_reason == 'completed') && (
+      (github.event.action == 'closed' && contains(github.event.issue.labels.*.name, 'user docs')) ||
       (github.event.action == 'labeled' && github.event.label.name == 'user docs'))
```
colin-rogers-dbt, line 35: do we need to emulate what we're doing in other places? https://github.com/dbt-labs/dbt-redshift/pull/847/files#diff-9b299312ba5915fe644f25c3ea9d363d232b089664545518e8e020720cc21baeR35

colin-rogers-dbt: okay on re-read I think we can just delete line 37 right?

McKnight-42: @mikealfare  any thoughts on this, I believe this was added to cover your suggestion that we may also want to handle instances of labels being able to kick up docs issue before in case the context is needed to be built up over time instead of after completion.

mikealfare: If we're only catching the close event (`github.event.action == 'closed'`) then we should update the `types` to only list `closed` instead of `[closed, labeled]`. If we also want to support adding the label after the issue is closed, then we need to keep the labeled event in `types` and update the logic here for that scenario. The fastest would be to remove `labeled` from `types` and leave that as a follow-up when/if we want to do it.

colin-rogers-dbt: Ok I just misunderstood the logic here, I think we should support adding a label after the issue is closed. This is the kind of papercut we'll just never get back to otherwise

mikealfare (approved):

McKnight-42:
@cla-bot check

PR #1276, opened 2024-07-10
automate internal release workflow.

McKnight-42:
https://github.com/dbt-labs/adapters-team-private/issues/168

### Problem
manually having to kick off the internal release workflow is a time eater, we can try to automate it as much as possible and just check it in the start of day.

### Solution
add a cron job call for early mornings on a specified day a week (currently weds) and have the workflow first compare commit to last release if there is a new commit continue with release else don't.

### Checklist

### DIFF
--- a/.github/workflows/release-internal.yml
+++ b/.github/workflows/release-internal.yml
@@ -54,6 +54,12 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # Fetch all history for all branches and tags
+
+      - name: Fetch ref
+        run: |
+          git fetch origin ${{ inputs.ref }}:${{ inputs.ref }}
 
       - name: Get last successful Release to Cloud run details and check for new commits
         id: check_new_commits
--- a/.github/workflows/release-internal.yml
+++ b/.github/workflows/release-internal.yml
@@ -8,7 +8,7 @@
 #
 # When?
 #
-# Manual trigger.
+# Manual trigger or cron job every wednesday morning.
 
 name: "Release to Cloud"
 run-name: "Release to Cloud off of ${{ inputs.ref }}"
@@ -32,20 +32,45 @@ on:
         required: true
         default: false
 
+  schedule:
+    - cron: '0 13 * * 3' # Every Wednesday at 8:00 AM central time
+
 defaults:
   run:
     shell: bash
 
 jobs:
+  get-latest-release:
+    runs-on: ubuntu-latest
+    outputs:
+      latest_release: ${{ steps.get_latest_release.outputs.latest_release }}
+    steps:
+      - name: Get latest release
+        id: get_latest_release
+        run: |
+          latest_release=$(curl --silent "https://api.github.com/repos/${{ github.repository }}/releases/latest" | jq -r .tag_name)
+          echo "::set-output name=latest_release::$latest_release"
+
+  check-for-new-commits:
+    runs-on: ubuntu-latest
+    needs: get-latest-release
+    outputs:
+      new_commits: ${{ steps.check_new_commits.outputs.new_commits }}
+    steps:
+      - name: Check for new commits since last release
+        id: check_new_commits
+        run: |
+          last_release_date=$(curl --silent "https://api.github.com/repos/${{ github.repository }}/releases/tags/${{ needs.get-latest-release.outputs.latest_release }}" | jq -r .published_at)
+          new_commits=$(git rev-list --count --since="$last_release_date" ${{ inputs.ref }})
+          echo "::set-output name=new_commits::$new_commits"
+
   invoke-reusable-workflow:
+    if: needs.check-for-new-commits.output.new_commits != '0'
     name: "Build and Release Internally"
-
     uses: "dbt-labs/dbt-release/.github/workflows/internal-archive-release.yml@main"
-
     with:
       package_test_command: "${{ inputs.package_test_command }}"
       dbms_name: "bigquery"
       ref: "${{ inputs.ref }}"
       skip_tests: "${{ inputs.skip_tests }}"
-
     secrets: "inherit"
--- a/.github/workflows/release-internal.yml
+++ b/.github/workflows/release-internal.yml
@@ -40,27 +40,28 @@ defaults:
     shell: bash
 
 jobs:
-  get-latest-release:
+  get-last-successful-run:
     runs-on: ubuntu-latest
     outputs:
-      latest_release: ${{ steps.get_latest_release.outputs.latest_release }}
+      commit_sha: ${{ steps.get_last_successful_run.outputs.commit_sha }}
     steps:
-      - name: Get latest release
-        id: get_latest_release
+      - name: Get last successful run details
+        id: get_last_successful_run
         run: |
-          latest_release=$(curl --silent "https://api.github.com/repos/${{ github.repository }}/releases/latest" | jq -r .tag_name)
-          echo "::set-output name=latest_release::$latest_release"
+          last_run_url=$(curl --silent -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" "https://api.github.com/repos/${{ github.repository }}/actions/runs?event=workflow_dispatch&status=success&branch=${{ github.ref }}" | jq -r '.workflow_runs[0].url')
+          commit_sha=$(curl --silent -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" $last_run_url | jq -r '.head_sha')
+          echo "::set-output name=commit_sha::$commit_sha"
 
   check-for-new-commits:
     runs-on: ubuntu-latest
-    needs: get-latest-release
+    needs: get-last-successful-run
     outputs:
       new_commits: ${{ steps.check_new_commits.outputs.new_commits }}
     steps:
       - name: Check for new commits since last release
         id: check_new_commits
         run: |
-          last_release_date=$(curl --silent "https://api.github.com/repos/${{ github.repository }}/releases/tags/${{ needs.get-latest-release.outputs.latest_release }}" | jq -r .published_at)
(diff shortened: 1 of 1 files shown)


.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,45 @@ on:
...
+    outputs:
+      new_commits: ${{ steps.check_new_commits.outputs.new_commits }}
+    steps:
+      - name: Check for new commits since last release
+        id: check_new_commits
+        run: |
+          last_release_date=$(curl --silent "https://api.github.com/repos/${{ github.repository }}/releases/tags/${{ needs.get-latest-release.outputs.latest_release }}" | jq -r .published_at)
+          new_commits=$(git rev-list --count --since="$last_release_date" ${{ inputs.ref }})
+          echo "::set-output name=new_commits::$new_commits"
+
```
McKnight-42, line 66: do we need to point somewhere else for this? won't internal releases be far ahead of the version releases? can we look at previous runs of this workflow and grab the commit from them?

McKnight-42: have updated workflow from this initial comment would love thoughts.

.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,46 @@ on:
...
+  check-for-new-commits:
+    runs-on: ubuntu-latest
+    needs: get-last-successful-release-run
+    outputs:
+      new_commits: ${{ steps.check_new_commits.outputs.new_commits }}
+    steps:
+      - name: Check for new commits since last release
+        id: check_new_commits
+        run: |
+          last_release_date=$(curl --silent "https://api.github.com/repos/${{ github.repository }}/releases/tags/${{ needs.get-last-successful-run.outputs.commit_sha }}" | jq -r .published_at)
```
colin-rogers-dbt, line 64: I think we want to retrieve commits not releases (i.e. we may not have a github release yet). Looks like this could work: 
```https://api.github.com/repos/dbt-labs/dbt-bigquery/commits/2c77893d0fb7b5d76ba85cc16712b9563a60329b```

File: .github/workflows/release-internal.yml [hunk 3]
```diff
@@ -32,20 +32,46 @@ on:
...
 jobs:
+  get-last-successful-release-run:
+    runs-on: ubuntu-latest
+    outputs:
+      commit_sha: ${{ steps.get_last_successful_release_run.outputs.commit_sha }}
+    steps:
+      - name: Get last successful Release to Cloud run details
+        id: get_last_successful_release_run
+        run: |
+          last_run_url=$(curl --silent -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" "https://api.github.com/repos/${{ github.repository }}/actions/runs?event=workflow_dispatch&status=success&branch=${{ github.ref }}&workflow_file=release.internal.yml" | jq -r '.workflow_runs[0].url')
```
colin-rogers-dbt, line 51: this is pretty clever

.github/workflows/release-internal.yml
McKnight-42: updating this more this morning i'm starting to wonder if this should be its own action that triggers our current workflow to supply the fields we need that way. thoughts?

McKnight-42:
question: should this be just releasing automatically off main or for every "supported" version

.github/workflows/release-internal.yml
mikealfare: I think the file should be `release-internal.yml` instead of `release.internal.yml`.

.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,56 @@ on:
...
+          echo "Last run URL: $last_run_url"
+          commit_sha=$(curl --silent -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" $last_run_url | jq -r '.head_sha')
+          echo "Commit SHA: $commit_sha"
+          echo "::set-output name=commit_sha::$commit_sha"
```
mikealfare, line 55: I think this needs to be the new syntax that uses `>> GITHUB_OUTPUT`. I forget what it is off the top of my head, but it's something like `echo "set commit_sha=$commit_sha" >> GITHUB_OUTPUT`.

.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,55 @@ on:
...
+    steps:
+      - name: Check for new commits since last Release to Cloud run
+        id: check_new_commits
+        run: |
+          last_commit_sha=${{ needs.get-last-successful-release-run.outputs.commit_sha }}
+          echo "Last successful commit SHA: $last_commit_sha"
+          latest_commit_sha=$(git rev-parse ${{ inputs.ref }})
+          echo "Latest commit SHA: $latest_commit_sha"
+          if [ "$last_commit_sha" != "$latest_commit_sha" ]; then
+            echo "::set-output name=has_new_commits::true"
```
mikealfare, line 71: Same as above, I think these need to use `>> GITHUB_OUTPUT`.

.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,55 @@ on:
...
+    outputs:
+      latest_commit_sha: ${{ steps.check_new_commits.outputs.latest_commit_sha }}
+      has_new_commits: ${{ steps.check_new_commits.outputs.has_new_commits }}
+    steps:
+      - name: Check for new commits since last Release to Cloud run
+        id: check_new_commits
+        run: |
+          last_commit_sha=${{ needs.get-last-successful-release-run.outputs.commit_sha }}
+          echo "Last successful commit SHA: $last_commit_sha"
+          latest_commit_sha=$(git rev-parse ${{ inputs.ref }})
```
mikealfare, line 68: Does this work if the repo is not checked out yet? This looks like it's using git locally and not the GH API, so I think you need to use a checkout step first.

.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,55 @@ on:
...
+          echo "Latest commit SHA: $latest_commit_sha"
+          if [ "$last_commit_sha" != "$latest_commit_sha" ]; then
```
mikealfare, line 70: I read this three times before realizing the difference. That could just be me, but maybe consider another name for `last_commit_sha`, something that indicates it's the last released commit.

.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,55 @@ on:
...
+            echo "::set-output name=has_new_commits::true"
+          else
+            echo "::set-output name=has_new_commits::false"
+          fi
+          echo "::set-output name=latest_commit_sha::$latest_commit_sha"
+
   invoke-reusable-workflow:
+    if: needs.check-for-new-commits.outputs.has_new_commits == 'true'
```
mikealfare, line 78: You need a `needs` node to be able to use `needs.check-for-new-commits`.

mikealfare (commented):
I think you're mostly there, just a few nits and catches.

mikealfare:
> question: should this be just releasing automatically off main or for every "supported" version

What do you mean by "supported" version? I think we're only ever releasing off `main`. And we do so as long as tests pass, which I guess means it's supported.

.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,55 @@ on:
...
+    runs-on: ubuntu-latest
+    outputs:
+      commit_sha: ${{ steps.get_last_successful_release_run.outputs.commit_sha }}
+    steps:
+      - name: Get last successful Release to Cloud run details
+        id: get_last_successful_release_run
+        run: |
+          last_run_url=$(curl --silent -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" "https://api.github.com/repos/${{ github.repository }}/actions/runs?event=workflow_dispatch&status=success&branch=main&workflow_file=release-internal.yml" | jq -r '.workflow_runs[0].url')
+          echo "Last run URL: $last_run_url"
+          commit_sha=$(curl --silent -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" $last_run_url | jq -r '.head_sha')
```
VersusFacit, line 53: Is there a way this can be `None`? Might be nice to validate and exit here if it's empty.

.github/workflows/release-internal.yml
```diff
@@ -32,20 +32,55 @@ on:
...
-
     uses: "dbt-labs/dbt-release/.github/workflows/internal-archive-release.yml@main"
-
     with:
       package_test_command: "${{ inputs.package_test_command }}"
       dbms_name: "bigquery"
       ref: "${{ inputs.ref }}"
       skip_tests: "${{ inputs.skip_tests }}"
-
-    secrets: "inherit"
```
VersusFacit, line 51: Is there a reason to change this? I thought we want things in quoted strings more for consistency

.github/workflows/release-internal.yml
```diff
@@ -8,7 +8,7 @@
 #
 # When?
 #
-# Manual trigger.
+# Manual trigger or cron job every wednesday morning.
```
VersusFacit, line 11: Can we describe the fact this job doesn't do anything if there are no new commits

.github/workflows/release-internal.yml
```diff
@@ -31,21 +32,76 @@ on:
...
+
+      - name: Fetch ref
+        run: |
+          git fetch origin ${{ inputs.ref }}:${{ inputs.ref }}
+
+      - name: Get last successful Release to Cloud run details and check for new commits
+        id: check_new_commits
+        run: |
+          # Get last successful Release to Cloud run details
+          last_run_url=$(curl --silent -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" "https://api.github.com/repos/${{ github.repository }}/actions/runs?event=workflow_dispatch&status=success&branch=main&workflow_file=release-internal.yml" | jq -r '.workflow_runs[0].url')
```
McKnight-42, line 68: this appears to be operating unexpected, i'm targeting workflows based on the the specific file but the returned object from the dry_run was for an `adapter integration test run`

[dry run](https://github.com/dbt-labs/dbt-bigquery/actions/runs/10046774056/job/27766989444)
[URL](https://api.github.com/repos/dbt-labs/dbt-bigquery/actions/runs/10042568588)

PR #1277, opened 2024-07-16
[CT-2819] default__alter_relation_add_remove_columns macro does not use quoting with case sensitive Snowflake relation 

McKnight-42:
https://github.com/dbt-labs/dbt-adapters/issues/250

### Problem
user raised issue around "While working with DBT incremental config: on_schema_change='append_new_columns'
The append new columns flag is not able to capture the correct case-sensitive column name and add it to the incremental table causing the run to fail."

they stated they were using snowflake.

### Solution
add a new test to dbt-adapters-tests to check that column quoting case sensitivity is expressed correctly, update all macros in adapters as needed if they do not use the default implementation and test default implementing macros to see if we need to update the dbt-adapters macro as well.

### Todo:
- [ ] test with rest of adapters (see if dbt-adapter default version of macro needs to be changed or any other adapter override version of macro).
- [ ] make sure new tests works in all adapters as needed
- [ ] update requirements files either dev-requirements or project.toml files before merger
### Checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,9 +3,7 @@
 git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
 git+https://github.com/dbt-labs/dbt-common.git
 git+https://github.com/dbt-labs/dbt-adapters.git@mcknight/ct-2819
-# -e /Users/matthewmcknight/git/dbt-adapters
-git+https://github.com/dbt-labs/dbt-adapters.git#subdirectory=dbt-tests-adapter
-# -e /Users/matthewmcknight/git/dbt-adapters/dbt-tests-adapter
+git+https://github.com/dbt-labs/dbt-adapters.git@mcknight/ct-2819#subdirectory=dbt-tests-adapter
 # if version 1.x or greater -> pin to major version
 # if version 0.x -> pin to minor
 black>=24.3
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,7 +3,7 @@
 git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
 git+https://github.com/dbt-labs/dbt-common.git
 git+https://github.com/dbt-labs/dbt-adapters.git@mcknight/ct-2819
-git+https://github.com/dbt-labs/dbt-adapters.git@mcknight/ct-2819#subdirectory=dbt-tests-adapterch
+git+https://github.com/dbt-labs/dbt-adapters.git@mcknight/ct-2819#subdirectory=dbt-tests-adapter
 # if version 1.x or greater -> pin to major version
 # if version 0.x -> pin to minor
 black>=24.3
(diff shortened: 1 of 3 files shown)


McKnight-42:
since bigquery naturally accepts case sensitiviy without needed quoting I made a test for that.

use cases I've seen that would require a column to be quoted are if a space or special characters are needed but trying to add those thus far has lead to failure for string literals in the model.

suggestions?

tests/functional/adapter/incremental/test_incremental_on_schema_change.py
```diff
@@ -251,3 +251,59 @@ def test_run_incremental_sync_all_columns_time_ingestion_partitioning(self, proj
...
+
+class BaseBigQueryIncrementalCaseSenstivityOnSchemaChange:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "src_artists.sql": _MODELS__SRC_ARTISTS,
+            "dim_artists.sql": _MODELS__DIM_ARTISTS,
+        }
+
+    def test_incremental_case_sensitivity(self, project):
```
mikealfare, line 290: I could be missing it, but this looks like the same test as in the base. Can we inherit from the new test case and just overwrite `models`? Also, this should just be a test class. We don't need the base class and the test class since this is in the concrete adapter.

tests/functional/adapter/incremental/test_incremental_on_schema_change.py
```diff
@@ -251,3 +251,59 @@ def test_run_incremental_sync_all_columns_time_ingestion_partitioning(self, proj
...
+    config(
+        materialized='table',
+    )
+}}
+{% if var("version", 0) == 0 %}
+    select {{ dbt.current_timestamp() }} as inserted_at, 'john' as name
+{% else %}
+    -- add a non-zero version to the end of the command to get a different version:
+    -- --vars "{'version': 1}"
+    select {{ dbt.current_timestamp() }} as inserted_at, 'engineer' as Job,
```
mikealfare, line 267: Is `Name` purposely skipped here? I noticed you had it in the `dbt-tests-adapter` version.

dev-requirements.txt
```diff
@@ -2,8 +2,10 @@
 # TODO: how to automate switching from develop to version branches?
 git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
 git+https://github.com/dbt-labs/dbt-common.git
-git+https://github.com/dbt-labs/dbt-adapters.git
+git+https://github.com/dbt-labs/dbt-adapters.git@mcknight/ct-2819
+# -e /Users/matthewmcknight/git/dbt-adapters
 git+https://github.com/dbt-labs/dbt-adapters.git#subdirectory=dbt-tests-adapter
```
mikealfare, line 7: You need to update the branch for `dbt-tests-adapter` too, to pick up the new test case.

tests/functional/adapter/incremental/test_incremental_on_schema_change.py
McKnight-42: yeah for bigquery when ran it seemed to count the column as already existing and would fail. removing and running seems fine even if I changed the casing of name to Name it still failed. for column already existing happy to look into it further.

my thought was that since bigquery checks column casing natively it sees them as the same thing so nothing new to add and that seems to borq the query

tests/functional/adapter/incremental/test_incremental_on_schema_change.py
McKnight-42: I had pulled the dbt-adapter ones into its own base case and test class so that the location I had them originally as part of `BaseIncrementalOnSchemaChangeSetup` but didn't want to always pull it in since bigquery requires its own default version

PR #1284, opened 2024-07-18
Pin `setuptools` to address setuptools#4483

mikealfare:
### Problem

`setuptools` released `71.0.0` and `71.0.1` last night which contains a breaking change. This is blocking our CI. See https://github.com/pypa/setuptools/issues/4483 for more information.

### Solution

- create a separate `build-requirements.txt` file to isolate build dependencies
- pin `setuptools~=70.0` in this file
- reference this file in `dev-requirements.txt` for backwards compatibility
- install from this file in workflows instead of running `pip install --upgrade setuptools ...`

### Checklist

### DIFF
--- a/build-requirements.txt
+++ b/build-requirements.txt
@@ -1,0 +1,5 @@
+bumpversion~=0.6.0
+check-wheel-contents~=0.6.0
+setuptools~=70.0
+twine~=5.0
+wheel~=0.42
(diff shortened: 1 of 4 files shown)


build-requirements.txt
```diff
@@ -0,0 +1,5 @@
+bumpversion~=0.6.0
+check-wheel-contents~=0.6.0
+setuptools~=70.0
+twine~=5.0
```
mikealfare, line 4: We were installing this directly and upgrading, so we never actually used 4.0. 4.0 is also broken, see https://github.com/pypa/twine/issues/977. From the post in there, the date of the fix, and the dates of the releases, the fix was released as part of 5.0.

mikealfare:
We don't actually need this.

PR #1291, opened 2024-07-18
Move code quality deps to precommit

mikealfare:
### Problem

We wind up with a ton of dependabot updates, most of them for linting dependencies. We also have two versions of linting config, one in `.pre-commit-config.yaml` and one in separate config files. Since we generally run linting via `pre-commit`, the latter doesn't get used.

### Solution

- move all linting dependencies and config into `.pre-commit-config.yaml`
- make sure linting calls use `pre-commit run` instead of calling directly (the tools will no longer be in the virtual environment)
- make linting updates resulting from `pre-commit run --all-files`

### Checklist


VersusFacit (approved):
Love the cleanup

PR #1346, opened 2024-09-12
Isolating distribution testing

### LINKED ISSUE #1290: Update the build check workflow
### Current Behavior

We install the wheel, check that the package is installed, then immediately install the sdist, and check that the package is installed again.

### Expected Behavior

We should be validating the packages in isolated environments. This arose from the `setuptools` / `packaging` issue for `dbt-core<1.4`, but it's something we should do for current versions as well. See https://github.com/dbt-labs/dbt-core/issues/10465 for the analogous `dbt-core` issue.

leahwicz:
resolves #1290 

### Problem
We validate wheel and source distributions in the same environment during the build validation step. This led to issues documented here https://github.com/dbt-labs/dbt-core/issues/10465. 

### Solution
This is the same solution as what we did in `dbt-snowflake` first: https://github.com/dbt-labs/dbt-snowflake/pull/1161
This is to bring parity to the other repos as well.

This is a test run of the Action running with the installation and testing of wheels and source distributions split into separate Actions Jobs that isolate them from one another: https://github.com/leahwicz/dbt-bigquery/actions/runs/10821506800/job/30023648113

### Checklist


.github/workflows/main.yml
```diff
@@ -176,17 +176,20 @@ jobs:
...
     steps:
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+
       - name: Install python dependencies
         run: |
           python -m pip install --user --upgrade pip
-          python -m pip install --upgrade wheel setuptools twine check-wheel-contents
```
leahwicz, line 188: These packages aren't needed but were copy and pasted from the above step I'm guessing. We are still able to install and run the distributions without issue (verified in the test run I linked in the description)

.github/workflows/main.yml
mikealfare, line 1: @leahwicz I accidentally reverted the matrix entry here and I can't update directly. Do you mind adding it back:

```yaml
# 179 
        dist-type: ["whl", "gz"]
```

PR #1388, opened 2024-10-29
use "direct" write for non-partitioned python model materializations

### LINKED ISSUE #1318: [Regression] Python model build failed when Google bucket retention policy is set (starting from dbt version 1.7)

### Current Behavior

Starting dbt version 1.7, when bucket retention policy is set, python model build will throw error Unhandled error while executing target/run/restream_bi/models/python/python_model_test.py[0m Google Cloud Dataproc Agent reports job failure. 

### Expected Behavior

In dbt version1.6, the python model does not get a build error even when when bucket retention policy is set.

### Steps To Reproduce

1. Change dbt version to 1.7 or later in Dev env
2. Set a value to bucket retention policy in bucket details in Google Cloud
3. Run `dbt run -s <python model>`

### Relevant log output

```shell
Google Cloud Dataproc Agent reports job failure.

Using the default container image
Waiting for container log creation
PYSPARK_PYTHON=/opt/dataproc/conda/bin/python
JAVA_HOME=/usr/lib/jvm/temurin-11-jdk-amd64
SPARK_EXTRA_CLASSPATH=
:: loading settings :: file = /etc/spark/conf/ivysettings.xml
/usr/lib/spark/python/lib/pyspark.zip/pyspark/pandas/__init__.py:49: UserWarning: 'PYARROW_IGNORE_TIMEZONE' environment variable was not set. It is required to set this environment variable to '1' in both driver and executor sides if you use pyarrow>=2.0.0. pandas-on-Spark will set it for you but it does not work if there is a Spark context already launched.
```

### Environment

```markdown
- OS:macOS
- Python/dbt: 
Error in dbt cloud version 1.7 and versionless
OK in dbt cloud version 1.6
```

bigquery

### Additional Context

_No response_

colin-rogers-dbt:
resolves #1318

In order to support partitioned materializations in BQ python models we we switched from "direct" to "indirect" mode when writing model results in Dataproc back to BigQuery. As the naming implies "indirect" temporarily stages data in the provided GCS bucket. If a user has a retention policy on the bucket this will fail as the bucket won't allow Dataproc to delete these temp files as it goes. 

This PR sidesteps that issue to ensure backwards compatibility with <1.7 created models by using "direct" write when a partitioned config is not provided.  

Note: I have not added any new testing to cover the case where a user has set a retention policy. Ultimately I think this is an edge case we don't need to test against but we should document that a bucket retention policy cannot be used with a partitioned python model as a follow up

### Checklist

### DIFF
--- a/dbt/include/bigquery/macros/materializations/table.sql
+++ b/dbt/include/bigquery/macros/materializations/table.sql
@@ -113,23 +113,29 @@ else:
   msg = f"{type(df)} is not a supported type for dbt Python materialization"
   raise Exception(msg)
 
+# For writeMethod we need to use "indirect" if materializing a partitioned table
+# otherwise we can use "direct". Note that indirect will fail if the GCS bucket has a retention policy set on it.
+{%- if partition_config is not none %}
+      {%- set write_method = 'indirect' -%}
+{%- else %}
+      {% set write_method = 'direct' -%}
+{%- endif %}
+
 df.write \
   .mode("overwrite") \
   .format("bigquery") \
+  .option("writeMethod", "{{ write_method }}") \
+  .option("writeDisposition", 'WRITE_TRUNCATE') \
   {%- if partition_config is not none %}
-  .option("writeMethod", "indirect") \
   {%- if partition_config.data_type | lower in ('date','timestamp','datetime') %}
   .option("partitionField", "{{- partition_config.field -}}") \
   {%- if partition_config.granularity is not none %}
   .option("partitionType", "{{- partition_config.granularity| upper -}}") \
   {%- endif %}
   {%- endif %}
-  {% else %}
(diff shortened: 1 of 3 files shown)


dbt/include/bigquery/macros/materializations/table.sql
```diff
@@ -113,10 +113,19 @@ else:
   msg = f"{type(df)} is not a supported type for dbt Python materialization"
   raise Exception(msg)
 
+# For writeMethod we need to use "indirect" if materializing a partitioned table
+# otherwise we can use "direct". Note that indirect will fail if the GCS bucket has a retention policy set on it.
+{%- if partition_config is not none %}
```
VersusFacit, line 118: nit: I'd flip this because it just reads easier to have the user not balance a negative in their head

PR #1395, opened 2024-11-05
Add retry factory to consolidate retry strategies across dbt-bigquery

mikealfare:
### Problem

We handle retries in different ways across different classes and layers of the application. We regularly get bug reports related to retries not being properly handled.

### Solution

Dedicate a class specifically to building retry strategies and remove this logic from the execution code.

### Checklist


mikealfare (commented):
I left comments on each file to summarize my changes and provide some context for why I made the changes. Hopefully this helps.

dbt/adapters/bigquery/clients.py
```diff
@@ -0,0 +1,67 @@
...
+@Retry()  # google decorator. retries on transient errors with exponential backoff
+def job_controller_client(credentials: BigQueryCredentials) -> JobControllerClient:
+    return JobControllerClient(
+        credentials=google_credentials(credentials),
+        client_options=ClientOptions(api_endpoint=_dataproc_endpoint(credentials)),
+    )
+
+
+@Retry()  # google decorator. retries on transient errors with exponential backoff
+def batch_controller_client(credentials: BigQueryCredentials) -> BatchControllerClient:
```
colin-rogers-dbt, line 48:
```suggestion
def dataproc_batch_controller_client(credentials: BigQueryCredentials) -> BatchControllerClient:
```
should we differentiate between bigquery and dataproc clients?

dbt/adapters/bigquery/clients.py
```diff
@@ -0,0 +1,67 @@
...
+@Retry()  # google decorator. retries on transient errors with exponential backoff
+def storage_client(credentials: BigQueryCredentials) -> StorageClient:
+    return StorageClient(
+        project=credentials.execution_project,
+        credentials=google_credentials(credentials),
+    )
+
+
+@Retry()  # google decorator. retries on transient errors with exponential backoff
+def job_controller_client(credentials: BigQueryCredentials) -> JobControllerClient:
```
colin-rogers-dbt, line 40:
```suggestion
def dataproc_job_controller_client(credentials: BigQueryCredentials) -> JobControllerClient:
```

dbt/adapters/bigquery/clients.py
```diff
@@ -0,0 +1,67 @@
...
+    try:
+        return _bigquery_client(credentials)
+    except DefaultCredentialsError:
+        _logger.info("Please log into GCP to continue")
+        setup_default_credentials()
+        return _bigquery_client(credentials)
+
+
+@Retry()  # google decorator. retries on transient errors with exponential backoff
+def storage_client(credentials: BigQueryCredentials) -> StorageClient:
```
colin-rogers-dbt, line 32:
```suggestion
def gcs_storage_client(credentials: BigQueryCredentials) -> StorageClient:
```

dbt/adapters/bigquery/python_submissions.py
```diff
@@ -1,187 +1,163 @@
...
-        return dataproc_v1.JobControllerClient(
-            client_options=self.client_options, credentials=self.GoogleCredentials
-        )
 
-    def _get_cluster_name(self) -> str:
-        return self.parsed_model["config"].get(
-            "dataproc_cluster_name", self.credential.dataproc_cluster_name
-        )
+    def submit(self, compiled_code: str) -> Job:
+        _logger.info(f"Submitting cluster job to: {self._cluster_name}")
```
colin-rogers-dbt, line 68: nit: These can probably be `debug`, the user will know the cluster they are submitting to / will see it if it fails

dbt/adapters/bigquery/python_submissions.py
```diff
@@ -1,187 +1,163 @@
...
+
+        # submit the batch
+        operation = self._batch_controller_client.create_batch(request)
+
+        # wait for the batch to complete
+        response: Batch = operation.result(polling=self._polling_retry)
+
+        return response
+
+    def _batch(self) -> Batch:
```
colin-rogers-dbt, line 122: nit: 
```suggestion
    def _get_batch_job_config(self) -> Batch:
```

dbt/adapters/bigquery/clients.py
mikealfare: Is there a BQ batch controller client? Or perhaps there's a batch controller client from a tool other than BQ or DataProc? Otherwise it feels a little like extra information that just makes the name longer.

dbt/adapters/bigquery/clients.py
mikealfare: Same here, are there going to be other storage clients? If so, then sure.

dbt/adapters/bigquery/python_submissions.py
mikealfare: I'm fine with that. I'm also fine just removing it. The only reason it's here is because that's what the current process is doing.

dbt/adapters/bigquery/python_submissions.py
mikealfare: I think we get an actual Batch there, not just a config; though I suppose a Batch is actually a config itself. I'd also like to remove extra verbs from our method names. Get feels redundant here, to the point that this could even be a property if we wanted. But how about a compromise. I searched for different opinions and found [this](https://melevir.medium.com/python-functions-naming-the-algorithm-74320a18278d) one that I can get behind at an overall level (not just naming this method). I think it also addresses why I don't like `get`, which I frankly didn't realize why previously. So if we go with that, this should be `_create_batch(self) -> Batch`.

dbt/adapters/bigquery/clients.py
colin-rogers-dbt: it's certainly possible but I'd argue that it reduces mental overhead as you could confuse this with some sort of local file storage client

dbt/adapters/bigquery/python_submissions.py
colin-rogers-dbt: I'm good with `_create_batch`

dbt/adapters/bigquery/clients.py
colin-rogers-dbt: I think it's more that because I'm in dbt-bigquery I'm going to assume that all client's connect to bigquery so this explicitly invalidates that assumption

colin-rogers-dbt (approved):

dbt/adapters/bigquery/clients.py
mikealfare: Fair, so then should we go with `gcs_client` to avoid the "PIN number" scenario?

dbt/adapters/bigquery/retry.py
```diff
@@ -0,0 +1,149 @@
...
+class RetryFactory:
+
+    def __init__(self, credentials: BigQueryCredentials) -> None:
+        self._retries = credentials.job_retries or 0
+        self._job_creation_timeout = credentials.job_creation_timeout_seconds
+        self._job_execution_timeout = credentials.job_execution_timeout_seconds
+        self._job_deadline = credentials.job_retry_deadline_seconds
+
+    def create_job_creation_timeout(self, fallback: Optional[float] = None) -> Optional[float]:
+        return (
```
VersusFacit, line 54: These can be simplified using `_Minute` in the default value

dbt/adapters/bigquery/retry.py
```diff
@@ -0,0 +1,149 @@
...
+        if self._retries == 0:
+            return False
+
+        # count all errors
+        self._error_count += 1
+
+        # if the error is retryable, and we haven't breached the threshold, log and continue
+        if _is_retryable(error) and self._error_count <= self._retries:
+            _logger.debug(
+                f"Retry attempt {self._error_count} of { self._retries} after error: {repr(error)}"
```
VersusFacit, line 111: nit: WS

dbt/adapters/bigquery/retry.py
```diff
@@ -0,0 +1,149 @@
...
+        return Retry(
+            predicate=_BufferedPredicate(self._retries),
+            initial=_DEFAULT_INITIAL_DELAY,
+            maximum=_DEFAULT_MAXIMUM_DELAY,
+            deadline=self._job_deadline,
+            on_error=_create_reopen_on_error(connection),
+        )
+
+
+class _BufferedPredicate:
```
VersusFacit, line 90: nit: we talked about other names that convey allowed # of retries

dbt/adapters/bigquery/retry.py
```diff
@@ -0,0 +1,149 @@
...
+_SECOND = 1.0
+_MINUTE = 60 * _SECOND
+_HOUR = 60 * _MINUTE
+_DAY = 24 * _HOUR
+_DEFAULT_INITIAL_DELAY = _SECOND
+_DEFAULT_MAXIMUM_DELAY = 3 * _SECOND
+_DEFAULT_POLLING_MAXIMUM_DELAY = 10 * _SECOND
+
+
+_REOPENABLE_ERRORS = (
```
VersusFacit, line 30: If this only gets used once, let's move it down

dbt/adapters/bigquery/connections.py
```diff
@@ -558,89 +432,148 @@ def copy_bq_table(self, source, destination, write_disposition):
...
-                ", ".join(source_ref.path for source_ref in source_ref_array),
-                destination_ref.path,
-            ),
-            conn=conn,
-            fn=copy_and_results,
+        msg = 'copy table "{}" to "{}"'.format(
+            ", ".join(source_ref.path for source_ref in source_ref_array),
+            destination_ref.path,
+        )
+        with self.exception_handler(msg):
```
VersusFacit, line 439: this changes the exception handling so we should wrap that in the original retry predicate logic somehow

VersusFacit (approved):

PR #1408, opened 2024-11-20
Retry on 503

### LINKED ISSUE #682: [ADAP-498] [Bug] BQ does not retry on 503

### Current Behavior

Current if BigQuery returns a 503 error we do not retry even though BigQuery recommends that as the course of action.

### Expected Behavior

This is not a regression but rather an oversight -- 503 errors should be both retryable and reopenable since it indicates a transient unavailable condition in BigQuery

### Steps To Reproduce

Transient -- requires intermittent error from BQ.

### Relevant log output

_No response_

### Environment

```markdown
- dbt-core: all 
- dbt-bigquery: all
```

### Additional Context

_No response_

mikealfare:
resolves #682

### Problem

We were missing some retry scenarios that BigQuery added over time. We also were not retrying all client factories.

### Solution

- add retries to all client factories
- instead of replacing the building retryable errors, extend them with our additions

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/clients.py
+++ b/dbt/adapters/bigquery/clients.py
@@ -1,7 +1,10 @@
 from google.api_core.client_info import ClientInfo
 from google.api_core.client_options import ClientOptions
 from google.auth.exceptions import DefaultCredentialsError
-from google.cloud import bigquery, dataproc_v1, storage
+from google.cloud.bigquery import Client as BigQueryClient, DEFAULT_RETRY as BQ_DEFAULT_RETRY
+from google.cloud.dataproc_v1 import BatchControllerClient, JobControllerClient
+from google.cloud.storage import Client as StorageClient
+from google.cloud.storage.retry import DEFAULT_RETRY as GCS_DEFAULT_RETRY
 
 from dbt.adapters.events.logging import AdapterLogger
 
@@ -16,7 +19,7 @@ from dbt.adapters.bigquery.credentials import (
 _logger = AdapterLogger("BigQuery")
 
 
-def create_bigquery_client(credentials: BigQueryCredentials) -> bigquery.Client:
+def create_bigquery_client(credentials: BigQueryCredentials) -> BigQueryClient:
     try:
         return _create_bigquery_client(credentials)
     except DefaultCredentialsError:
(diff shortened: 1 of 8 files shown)


dbt/adapters/bigquery/clients.py
```diff
@@ -1,8 +1,7 @@
 from google.api_core.client_info import ClientInfo
 from google.api_core.client_options import ClientOptions
-from google.api_core.retry import Retry
 from google.auth.exceptions import DefaultCredentialsError
-from google.cloud.bigquery import Client as BigQueryClient
+from google.cloud.bigquery import Client as BigQueryClient, DEFAULT_RETRY
```
mikealfare, line 4: As it turns out, `Retry()` with no arguments doesn't really have defaults that we want, but this is what we were using for the BQ client factory. We weren't retrying at all on other factories.

dbt/adapters/bigquery/connections.py
mikealfare, line 1: These are all due to renaming retry methods to be more explicit.

dbt/adapters/bigquery/impl.py
mikealfare, line 1: The fallback was the same and we only call this here, so I pushed it back into the called method to simplify things.

dbt/adapters/bigquery/python_submissions.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, parsed_model: Dict, credentials: BigQueryCredentials) -> None
 
         # set retry policy, default to timeout after 24 hours
         retry = RetryFactory(credentials)
-        self._polling_retry = retry.create_polling(
+        self._polling_retry = retry.create_job_execution_polling(
```
mikealfare, line 46: Another renaming for the sake of being explicit.

tests/unit/test_bigquery_connection_manager.py
```diff
@@ -39,7 +39,7 @@ def setUp(self):
     def test_retry_connection_reset(self, mock_client_factory):
         new_mock_client = mock_client_factory.return_value
 
-        @self.connections._retry.create_reopen_with_deadline(self.mock_connection)
+        @self.connections._retry.create_job_execution_retry_with_reopen(self.mock_connection)
```
mikealfare, line 42: Renaming

dbt/adapters/bigquery/retry.py
```diff
@@ -1,10 +1,9 @@
 from typing import Callable, Optional
 
-from google.api_core.exceptions import Forbidden
 from google.api_core.future.polling import DEFAULT_POLLING
 from google.api_core.retry import Retry
-from google.cloud.bigquery.retry import DEFAULT_RETRY
-from google.cloud.exceptions import BadGateway, BadRequest, ServerError
+from google.cloud.bigquery.retry import DEFAULT_JOB_RETRY
```
mikealfare, line 5: In general, we're retrying jobs, we should use this default instead of `DEFAULT_RETRY`. We still use `DEFAULT_RETRY` for the client factories where it makes more sense.

dbt/adapters/bigquery/retry.py
```diff
@@ -1,10 +1,9 @@
...
+from google.cloud.exceptions import BadRequest
```
mikealfare, line 6: The other exceptions are already captured in the predicate on `DEFAULT_JOB_RETRY`. This is already tested in a unit test (and is how I learned we add this one ourselves).

mikealfare (commented):
I left a self-review for context. Also, given this whole PR is retrying transient errors, what is expected from a testing effort? We already have some unit tests that verify that we retry errors as expected, but they don't test them in the context of other calls. e.g. we don't verify that if we try to load a file into BQ and we get a 503, that we in fact retry. Would testing this be going too far?

mikealfare:
We should wait until #1431 is merged and rebase this on that. They address similar things and that PR covers a lot of what was initially implemented here.

dbt/adapters/bigquery/clients.py
```diff
@@ -28,23 +27,23 @@ def create_bigquery_client(credentials: BigQueryCredentials) -> BigQueryClient:
...
+@DEFAULT_RETRY
 def create_gcs_client(credentials: BigQueryCredentials) -> StorageClient:
     return StorageClient(
         project=credentials.execution_project,
         credentials=create_google_credentials(credentials),
     )
 
 
-@Retry()  # google decorator. retries on transient errors with exponential backoff
+@DEFAULT_RETRY
```
colin-rogers-dbt, line 38: Same as above but for DataProc

dbt/adapters/bigquery/clients.py
```diff
@@ -28,23 +27,23 @@ def create_bigquery_client(credentials: BigQueryCredentials) -> BigQueryClient:
         return _create_bigquery_client(credentials)
 
 
-@Retry()  # google decorator. retries on transient errors with exponential backoff
+@DEFAULT_RETRY
```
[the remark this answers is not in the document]
mikealfare: Good catch, I updated this.

dbt/adapters/bigquery/clients.py
mikealfare: Same here.

colin-rogers-dbt (approved):

PR #1423, opened 2024-12-05
Remove custom retry in get_table call

mikealfare:
resolves #

### Problem

We added a custom retry predicate in a call to `get_table` when adding the retry factory and it appears to be missing some retryable errors (versus the default).

### Solution

Remove the custom retry and use the built in default.

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/connections.py
+++ b/dbt/adapters/bigquery/connections.py
@@ -519,10 +519,7 @@ class BigQueryConnectionManager(BaseConnectionManager):
         # backwards compatibility: fill in with defaults if not specified
         database = database or conn.credentials.database
         schema = schema or conn.credentials.schema
-        return client.get_table(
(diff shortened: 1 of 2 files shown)


dbt/adapters/bigquery/connections.py
```diff
@@ -519,10 +519,7 @@ def get_bq_table(self, database, schema, identifier) -> Table:
         # backwards compatibility: fill in with defaults if not specified
         database = database or conn.credentials.database
         schema = schema or conn.credentials.schema
-        return client.get_table(
-            table=self.table_ref(database, schema, identifier),
-            retry=self._retry.create_reopen_with_deadline(conn),
```
mikealfare, line 524: This was added during the retry factory work in #1395 but should have remained unset to continue using the default. No test changes are needed as this is reverting a change.

colin-rogers-dbt (approved):

PR #1431, opened 2024-12-11
[Bug] Use bigquery default retryable exceptions

### LINKED ISSUE #263: [CT-1015] job_retries not obeyed when job_execution_timeout_seconds is exceeded
### Describe the bug

The `job_retries` BigQuery job configuration setting is not used in conjunction with `job_execution_timeout_seconds` setting. Instead of retrying a job that times out, dbt immediately fails with `Operation did not complete within the designated timeout.`

### Steps To Reproduce

Sure. First use the official dbt Docker image, and run `dbt init` to initialise a project:

```bash
docker build -t dbt-bigquery-issue-263 --no-cache -<<'EOF'
FROM ghcr.io/dbt-labs/dbt-bigquery:1.1.0
WORKDIR /tmp
RUN dbt init --skip-profile-setup issue_263
EOF
```

Make a `profiles.yml` file similar to this. Take special note of the `job_execution_timeout_seconds` value, which I've exaggerated to 1 second to force a query timeout to occur.  Also note that `job_retries=5` Fill in the `project`key.

```yml
issue_263:
  target: default
  outputs:
    default:
      type: bigquery
      method: service-account
      project: '<your project>'
      dataset: 'issue_263'
      threads: 1
      keyfile: /root/keyfile.json
      job_execution_timeout_seconds: 1
      job_retry_deadline_seconds: 200
      job_creation_timeout_seconds: 30
      job_retries: 5
      priority: interactive
      location: US
```

Then, run the built docker image, executing a dbt run on the example `my_first_dbt_model` model. Map your `profiles.yml` file, and `keyfile.json` into the container:

```bash
docker run  \
-v `pwd`/profiles.yml:/root/.dbt/profiles.yml \
-v `pwd`/keyfile.json:/root/keyfile.json dbt-bigquery-issue-263 run \
--project-dir /tmp/issue_263 -m my_first_dbt_model
``` 

### Expected behavior

The model `my_first_dbt_model` would fail, and retry five times.

After dbt errors/exits, the job continues running on BigQuery, and eventually succeeds:

image

### Additional Context

The retries *are* obeyed in the event of other BigQuery failures. For example, if you introduce an intentional syntax error in your model, it will retry up to `job_retries`.

### System information
```
Core:
  - installed: 1.1.0
  - latest:    1.2.0 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - bigquery: 1.1.0 - Update available!

  At least one plugin is out of date or incompatible with dbt-core.
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation
```

**The operating system you're using:** Using the official Docker Image...

```bash
root@7fbf851eb622:/# lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
```

**The output of `python --version`:** Python 3.10.3

mikealfare:
resolves #263

### Problem

We're not retrying in all scenarios that BigQuery recommends. This is causing unexpected behavior for users. This likely happened as a result of us maintaining our own list of retryable errors, which fell out of sync with BigQuery's development over time.

### Solution

Instead of maintaining our own list, we should rely on BigQuery's list of retryable errors and reasons. However, we have one scenario that is not covered by BigQuery's defaults, `BadRequest`. We need to also check this scenario after checking BigQuery's defaults to maintain backwards compatibility.

### Checklist

### DIFF
--- a/dbt/adapters/bigquery/retry.py
+++ b/dbt/adapters/bigquery/retry.py
@@ -3,8 +3,8 @@ from typing import Callable, Optional
 from google.api_core.exceptions import Forbidden
 from google.api_core.future.polling import DEFAULT_POLLING
 from google.api_core.retry import Retry
-from google.cloud.bigquery.retry import DEFAULT_RETRY
-from google.cloud.exceptions import BadGateway, BadRequest, ServerError
+from google.cloud.bigquery.retry import DEFAULT_RETRY, _job_should_retry
+from google.cloud.exceptions import BadRequest
 from requests.exceptions import ConnectionError
 
 from dbt.adapters.contracts.connection import Connection, ConnectionState
@@ -74,6 +74,11 @@ class _DeferredException:
         self._retries: int = retries
         self._error_count = 0
 
+    @staticmethod
+    def _is_retryable(error: Exception) -> bool:
+        """Return true for errors that are unlikely to occur again if retried."""
+        return _job_should_retry(error) or isinstance(error, BadRequest)
+
     def __call__(self, error: Exception) -> bool:
         # exit immediately if the user does not want retries
         if self._retries == 0:
@@ -83,7 +88,7 @@ class _DeferredException:
         self._error_count += 1
 
         # if the error is retryable, and we haven't breached the threshold, log and continue
-        if _is_retryable(error) and self._error_count <= self._retries:
+        if self._is_retryable(error) and self._error_count <= self._retries:
             _logger.debug(
                 f"Retry attempt {self._error_count} of {self._retries} after error: {repr(error)}"
             )
@@ -113,16 +118,3 @@ def _create_reopen_on_error(connection: Connection) -> Callable[[Exception], Non
                 raise FailedToConnectError(str(e))
 
     return on_error
-
-
-def _is_retryable(error: Exception) -> bool:
-    """Return true for errors that are unlikely to occur again if retried."""
-    if isinstance(
-        error, (BadGateway, BadRequest, ConnectionError, ConnectionResetError, ServerError)
-    ):
-        return True
-    elif isinstance(error, Forbidden) and any(
-        e["reason"] == "rateLimitExceeded" for e in error.errors
-    ):
-        return True
-    return False
--- a/dbt/adapters/bigquery/retry.py
+++ b/dbt/adapters/bigquery/retry.py
@@ -1,6 +1,5 @@
 from typing import Callable, Optional
 
-from google.api_core.exceptions import Forbidden
 from google.api_core.future.polling import DEFAULT_POLLING
 from google.api_core.retry import Retry
 from google.cloud.bigquery.retry import DEFAULT_RETRY, _job_should_retry
--- a/dbt/adapters/bigquery/retry.py
+++ b/dbt/adapters/bigquery/retry.py
@@ -3,7 +3,6 @@ from typing import Callable, Optional
 from google.api_core.future.polling import DEFAULT_POLLING
 from google.api_core.retry import Retry
 from google.cloud.bigquery.retry import DEFAULT_RETRY, _job_should_retry
-from google.cloud.exceptions import BadRequest
 from requests.exceptions import ConnectionError
 
 from dbt.adapters.contracts.connection import Connection, ConnectionState
@@ -73,10 +72,6 @@ class _DeferredException:
         self._retries: int = retries
         self._error_count = 0
 
-    @staticmethod
-    def _is_retryable(error: Exception) -> bool:
-        """Return true for errors that are unlikely to occur again if retried."""
-        return _job_should_retry(error) or isinstance(error, BadRequest)
 
     def __call__(self, error: Exception) -> bool:
         # exit immediately if the user does not want retries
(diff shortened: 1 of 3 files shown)


dbt/adapters/bigquery/retry.py
```diff
@@ -113,16 +117,3 @@ def on_error(error: Exception):
                 raise FailedToConnectError(str(e))
 
     return on_error
-
-
-def _is_retryable(error: Exception) -> bool:
```
mikealfare, line 118: I moved this into the class that calls it as a static method to make it more clear that this isn't used anywhere else. In fact, we wouldn't need this at all if we remove `BadRequest` as retryable since `_is_retryable == _job_should_retry` in that scenario.

dbt/adapters/bigquery/retry.py
```diff
@@ -113,16 +117,3 @@ def on_error(error: Exception):
...
-    """Return true for errors that are unlikely to occur again if retried."""
-    if isinstance(
-        error, (BadGateway, BadRequest, ConnectionError, ConnectionResetError, ServerError)
-    ):
-        return True
-    elif isinstance(error, Forbidden) and any(
-        e["reason"] == "rateLimitExceeded" for e in error.errors
```
mikealfare, line 125: This condition is caught in the default retryable reasons (for more than just `Forbidden`).

dbt/adapters/bigquery/retry.py
```diff
@@ -113,16 +117,3 @@ def on_error(error: Exception):
                 raise FailedToConnectError(str(e))
 
     return on_error
-
-
-def _is_retryable(error: Exception) -> bool:
-    """Return true for errors that are unlikely to occur again if retried."""
-    if isinstance(
-        error, (BadGateway, BadRequest, ConnectionError, ConnectionResetError, ServerError)
```
mikealfare, line 121: Only `BadRequest` is missing from BQ's defaults; the rest can be removed since they would be caught in `_job_should_retry`.

tests/unit/test_bigquery_connection_manager.py
```diff
@@ -53,7 +53,7 @@ def generate_connection_reset_error():
         assert new_mock_client is not self.mock_client
 
     def test_is_retryable(self):
-        _is_retryable = dbt.adapters.bigquery.retry._is_retryable
+        _is_retryable = dbt.adapters.bigquery.retry._DeferredException(1)._is_retryable
```
mikealfare, line 56: `_is_retryable` was moved from a function to a method on `_DeferredException`, otherwise this test should still work the same.

dbt/adapters/bigquery/retry.py
```diff
@@ -74,6 +73,11 @@ def __init__(self, retries: int) -> None:
         self._retries: int = retries
         self._error_count = 0
 
+    @staticmethod
+    def _is_retryable(error: Exception) -> bool:
```
mikealfare, line 77: If we decide to remove `BadRequest`, this should go away and we should call `_job_should_retry` directly. We may decide to remove `BadRequest` since we would likely expect all retries to fail as well given the error appears to be on the user's side (malformed request) rather than BQ's side (e.g. resource unavailable).

dbt/adapters/bigquery/retry.py
```diff
@@ -1,10 +1,9 @@
 from typing import Callable, Optional
 
-from google.api_core.exceptions import Forbidden
 from google.api_core.future.polling import DEFAULT_POLLING
 from google.api_core.retry import Retry
-from google.cloud.bigquery.retry import DEFAULT_RETRY
-from google.cloud.exceptions import BadGateway, BadRequest, ServerError
+from google.cloud.bigquery.retry import DEFAULT_RETRY, _job_should_retry
```
mikealfare, line 5: While this is a private function, there is no reasonable way to capture BQ's defaults here. We need to add an exception into list that goes into this function, which would then require vendoring the whole function. Even if we remove `BadRequest`, we would then need to pull a private attribute (`_predicate`) off of `DEFAULT_JOB_RETRY` so that we can implement the retry cap (`job_retries`).

dbt/adapters/bigquery/retry.py
mikealfare: We determined we can remove `BadRequest` here. The rationale is that we would expect a bad request to always be a bad request since the issue originates from the user side and not the BQ side.

tests/unit/test_bigquery_connection_manager.py
```diff
@@ -65,7 +65,7 @@ def test_is_retryable(self):
         service_unavailable_error = exceptions.ServiceUnavailable("service is unavailable")
 
         self.assertTrue(_is_retryable(internal_server_error))
-        self.assertTrue(_is_retryable(bad_request_error))
+        self.assertFalse(_is_retryable(bad_request_error))  # this was removed after initially being included
```
mikealfare, line 68: It's worth keeping this, but flipping it, to be clear that this was once supported, and no longer is supported.

tests/unit/test_bigquery_connection_manager.py
```diff
@@ -65,7 +65,9 @@ def test_is_retryable(self):
         service_unavailable_error = exceptions.ServiceUnavailable("service is unavailable")
 
         self.assertTrue(_is_retryable(internal_server_error))
-        self.assertTrue(_is_retryable(bad_request_error))
+        self.assertFalse(
```
mikealfare, line 68: It's worth keeping this, but flipping it, to be clear that this was once supported, and no longer is supported.

VersusFacit (approved):

